PR #2, opened 2021-05-10
Handle wikilinks

srid:
- [x] Consider using https://github.com/jgm/commonmark-hs/blob/master/commonmark-extensions/src/Commonmark/Extensions/Wikilinks.hs 
- [x] https://github.com/srid/ema/pull/27


srid:
Can't use the upstream extension https://github.com/jgm/commonmark-hs/issues/69 because it

a) doesn't handle flipped position of custom text
b) doesn't handle folgezettel `#` prefix/suffix

jgm could be convinced about (a), but I dont believe he will accept (b) - since that's neuron specific.

srid:
https://github.com/jgm/commonmark-hs/issues/69#issuecomment-836084651

PR #3, opened 2021-05-10
Switch to Heist

srid:
Heist, although not popular, appears to be more safer. And the performance is reasonable for static site generation.﻿

- [x] proof of concept
- [x] context vars and make site work
- [x] remove old code
- [x] publish microblog


srid:
https://notes.srid.ca/heist-start

PR #31, opened 2021-05-26
Support --override-index-data argument

### LINKED ISSUE #20: Allow overriding base URL in CLI
Emanote's default template hardcodes `<base href="/">` via reading the "/" from `index.yaml` (`templates.baseUrl: /`). The default works in live server. But its value differs in certain deployment environments, like GitHub Pages. We should provide a way to override this when generating the static HTML (i.e., outside of live server).

```sh
emanote --override-index-data='{"template": {"baseUrl": "srid.github.io/mynotes"} }' gen
```

Note that this command itself is general, and allows overriding anything in the "index data" (which currently is represented by the top-level `index.yaml`). We can leverage the existing `Data.Aeson.Extra.Merge` to do this.

Goal: continue to use `/` as baseUrl in live server mode, but something else during deployment.

hi-learn-triz-and-usit:
Fixes #20

### DIFF
--- a/emanote.cabal
+++ b/emanote.cabal
@@ -51,7 +51,8 @@ executable emanote
     , monad-logger
     , mtl
     , neat-interpolation
-    , pandoc-link-context   >=1.0.2
+    , optparse-applicative
+    , pandoc-link-context  >=1.0.2
     , pandoc-types
     , profunctors
     , relude
@@ -91,6 +92,7 @@ executable emanote
   autogen-modules:    Paths_emanote
   other-modules:
     Emanote.Class
+    Emanote.Cli
     Emanote.Logging
     Emanote.Model
     Emanote.Model.Meta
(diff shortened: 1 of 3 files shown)


hi-learn-triz-and-usit:
New help looks like this
```
Emanote

Usage: emanote [--override-index-data JSON] [-C PATH] [COMMAND]
  Emanote - A spiritual successor to Neuron

Available options:
  --override-index-data JSON
                           Override the default parameters in index.yaml. For
                           example, to set the baseUrl, use {"template":
                           {"baseUrl": URL}}.
  -C PATH                  Run as if ema was started in PATH instead of the
                           current working directory.
  -h,--help                Show this help text

Available commands:
  gen                      Generate static HTML files
```

hi-learn-triz-and-usit:
The actual functionality is not yet hooked up.

hi-learn-triz-and-usit:
With #20 closed, what do you want done to support #35

srid:
Implementations details in https://github.com/srid/emanote/issues/20#issuecomment-848819699 - but the comment below has an open question; I haven't had any chance to think about that yet, unless you already have ideas on that.

hi-learn-triz-and-usit:
It seems like I won't have time to work on even this simple issue 😞 . Oh, well. Closing so as not to prevent others from taking it over.

Issue #35, opened 2021-05-27
Allow specifying additional notebooks in CLI


srid:
Add a way to specify additional notebook paths (they are ultimately "merged" in order). 

- [x] Rethink Ema's `-C` argument (chdir behave makes no sense with multiple arguments)
- [x] Implement this issue

---
Proposed CLI:

```bash
# Does nothing; just prints --help
emanote

# Fails, because requires a path argument
emanote run

# Runs emanote live server on the given folder (the only layer)
emanote run .

# Runs emanote gen command, with the given layers in argument order
emanote gen ./.deploy/github ./

# Run emanote live server on current directory, using common themes and public notes (3 layers)
cd ~/MyPrivateNotes
emanote run ~/myemanotethemes ~/MyPublicNotes ./
```

_Originally posted by @srid in https://github.com/srid/emanote/issues/20#issuecomment-849043533_

---

Implementation:

Consider extending `fsLayers` based on the paths passed in CLI,

https://github.com/srid/emanote/blob/78c9a2cf39d20fd90015e810b6d244cc270df49e/src/Emanote/Source.hs#L22-L24

This may involve replacing `LocUser` with a `LocUserProvided FilePath`

https://github.com/srid/emanote/blob/78c9a2cf39d20fd90015e810b6d244cc270df49e/src/Emanote/Source/Loc.hs#L7-L15

hi-learn-triz-and-usit:
Why do we prefer files from the right over files from the left? The main functional advantage of that comes up when programmatically constructing the command, but I don't think that's a likely situation; otherwise it's a matter of perspective and aesthetics.

srid:
It's the more natural order to me (you stack 'em up on top of another as you read left-to-right; plus the 'common' stuff gets put ahead, and you can create shell aliases easily that way), and I doubt most users would find it non-intuitive (but if they did, it can be changed).

srid:
contd. from https://github.com/srid/emanote/issues/20#issuecomment-848824175 

> ```
> LocUser (NonEmpty FilePath)
> ```
> 
> Actually it should just be `LocUser FilePath`. However, we break the order assumption ... hmm gotta think.
> 
> EDIT: some options available,
> 
> * use `LocUser Int FilePath` (where the Ind acts as an index), but that is hackish.
> * use https://hackage.haskell.org/package/ordered-containers-0.2.2/docs/Data-Set-Ordered.html

[...]

PR #41, opened 2021-05-28
Overlay configuration proposal

hi-learn-triz-and-usit:
I tried to follow the GHC proposal template, but I got a little burned out at the end.

[Rendered](https://github.com/hi-learn-triz-and-usit/emanote/blob/proposal-1/docs/proposals/000-overlays-via-configs.md)

### DIFF
--- a/docs/proposals/000-overlays-via-configs.md
+++ b/docs/proposals/000-overlays-via-configs.md
@@ -1,0 +1,88 @@
+# Use config files in top level directory to specify overlays
+
+Emanote has a powerful overlay system which enables different directories to be merged; this enables
+sharing data and configuration across projects. This functionality is not yet exposed to the user.
+
+## Motivation
+
+Exposing this functionality to the user addresses the following use cases.
+
+### Configuration overrides
+
+Suppose I have a site which uses a different base url in development than in production. Overlaying
+performs a structural merge on `.yaml` files, so, supposing `dev` is a directory that contains an
+`index.yaml` with development-specific parameters and `prod` contains the site with the production
+parameters, let's refer to the result of overlaying `dev` on top of `prod` as `prod+dev`. Running
+`emanote gen` on `prod` produces HTML ready to deploy to production; running `emanote gen` on
+`prod+dev` produces HTML ready to deploy to development. Notably, `prod` required no changes, and
+`dev` required almost no configuration.
+
+This proposal discusses how the user should represent things like `prod+dev`.
+
+### Composition
+
+It is relatively common to want to develop a site that builds upon another, or conversely to
+develop a site (or component thereof) that is meant to be built upon.
+
+#### Theming
+
+- I want to publish a theme, but I want users to be able to customize it.
+- I use `emanote` to develop multiple sites, but I want a consistent theme across all of them.
+
+#### Content
+
+- I want to have a public notes repository, but I also want a private one, which has more connections.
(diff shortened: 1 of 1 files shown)


srid:
There are a couple interesting things this proposal adds, or facilitate adding, on top, namely:

- Ability to compose notebooks automatically including their layers 
- Ability to support "remote notebooks" (eg: `github:srid/emanote-themes` to pull in themes automatically, maybe via nix flakes or direct git clone, and then using its layers automatically).
  - I prefer Nix but something like .gitmodules would also work. Reproducibility is important.

---

Note also these related feature requests in neuron (which emanote's notebook layers feature automatically addresses):

- https://github.com/srid/neuron/pull/339
- https://github.com/srid/neuron/discussions/222#discussioncomment-248875

---

Initial verdict on this RFC: CLI arguments (#35) are good enough *for now*. If/when I reach a stage where I am thinking to myself that "I cannot live without composable layer configuration feature" (and I do not think that at all right now) and/or users start asking for it (with individual examples/ rationale to boot) more, then I would consider adding some well thought-out implementation of this RFC in emanote. Until then, specify the multiple-notebooks manually in the CLI (#35) and let one's usage over time dictate any further inputs to this RFC. Besides, the "merge" semantics for individual files types when using multiple layers are in their proto form right now (yaml is json-merged, but .md is replaced), and not fully fleshed out yet (I'm thinking daily notes should be merged). So that's another reason to consider revisiting this RFC much later in time.

srid:
>  the "merge" semantics for individual files types when using multiple layers [...] not fully fleshed out yet (I'm thinking daily notes should be merged)

That specific example of daily notes can be configured like this: take a list of pairs `(FilePattern, MergeSemantic)`, i.e., `[("Daily/**/*.md", Merge)]` and apply those merge semantics overriding emanote defaults (thus, all Markdown files will use the "Replace" semantic, except for daily notes which get merged). 

To support this in the CLI, we may imagine adding an option, eg: `emanote --merge-semantics='Daily/**/*.md:merge' path1 path2 ...`. Alternatively -- and this certainly would be more "cleaner" (esp. as we are no longer specifying merely a list of paths) as @hi-learn-triz-and-usit  argues -- they may be specified in the layer configuration file. Thus, the layer configuration file will contain not only a list of paths (local or remote), but also the configuration for how to merge individual file types (overriding emanote's default).

Open questions: 

- are there any other kinds of things that can (only) be configured at the layer configuration level?
- Like the aforementioned daily notes example, what other relevant real-world use-cases exist? I guess we'll know over time.

srid:
> are there any other kinds of things that can (only) be configured at the layer configuration level?

Another thing that belongs in layer configuration:

- File ignore patterns
  - The default is to ignore all `.*` paths (dotfiles), but we might want to let the user customize that.
  - As an aside (and I haven't thought this in detail), this might be subsumed under the aforementioned 'merge semantics' (where the semantic here ==  "ignore")
    - That said, merge semantics are one level higher than layers, whereas ignore pattern could be configured at individual layer level (unless we enforce it at the global level, which however doesn't make much sense)

@hi-learn-triz-and-usit Okay, so I'm more convinced that your overlay configuration is a worthy concept to implement. But it should remain optional. And to be implemented on a 'as needed' basis.

srid:
Since this is linked in #37 (and can be reviewed at any time latter), I'll close this PR for now, as it hasn't seen any recent activity.

Issue #70, opened 2021-06-19
Allow setting title in YAML frontmatter


srid:
Note title is inferred from,

1. Level-1 Markdown heading if used, otherwise
2. Filename

In addition, support `title` attribute from the Markdown's YAML frontmatter too, so that it overrides the above two.

Change should possibly happen here: 

https://github.com/srid/emanote/blob/8fb4aa8cfaf097a61b474365a96b19b512a363c2/src/Emanote/Model/Note.hs#L100-L103

srid:
Done in #71

PR #71, opened 2021-06-19
Allow setting title in YAML frontmatter

carliros:
https://github.com/srid/emanote/issues/70

### DIFF
--- a/src/Emanote/Model/Note.hs
+++ b/src/Emanote/Model/Note.hs
@@ -102,7 +102,7 @@ noteTitle note@Note {..} =
   let yamlNoteTitle = lookupMeta (one "title") note
       fileNameTitle = R.routeBaseName . R.lmlRouteCase $ _noteRoute
       notePandocTitle = getPandocTitle _noteDoc
-   in fromMaybe (fromMaybe fileNameTitle notePandocTitle) yamlNoteTitle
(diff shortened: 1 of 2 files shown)


src/Emanote/Model/Note.hs
```diff
@@ -98,9 +98,11 @@ lookupMeta k =
...
 
 noteTitle :: Note -> Text
-noteTitle Note {..} =
-  fromMaybe (R.routeBaseName . R.lmlRouteCase $ _noteRoute) $
-    getPandocTitle _noteDoc
+noteTitle note@Note {..} =
+  let yamlNoteTitle = lookupMeta (one "title") note
+      fileNameTitle = R.routeBaseName . R.lmlRouteCase $ _noteRoute
+      notePandocTitle = getPandocTitle _noteDoc
+   in fromMaybe (fromMaybe fileNameTitle notePandocTitle) yamlNoteTitle
```
srid: By the way, you can use [`Alternative`](https://hackage.haskell.org/package/base-4.15.0.0/docs/Control-Applicative.html#t:Alternative) to refactor this into,

```haskell
fromMaybe fileNameTitle $ yamlNoteTitle <|> notePandocTitle 
```

carliros: cool, learned that today!

Issue #69, opened 2021-06-19
mailto: links should not be handled like resource links


srid:
Use `foo@bar.com` in Markdown, and Emanote will render it as a broken link.

Fix: change emanote to be aware of `mailto:` protocol, and ignore handling it. 

Potential solution: add another `guard` in here: 

https://github.com/srid/emanote/blob/8fb4aa8cfaf097a61b474365a96b19b512a363c2/src/Emanote/Model/Link/Rel.hs#L80-L82

carliros:
I see this
```
-- TODO: Handle mailto: etc.
```
in WikiLinks.hs, line 53. 
I think we need to remove that, correct? Because that is being addressed in Rel.hs

Same for line 54. This is repetitive, correct?

srid:
Resolved by #72

> I think we need to remove that, correct? Because that is being addressed in Rel.hs

Ah, yes, you are right. Good catch.

PR #72, opened 2021-06-19
mailto: links should not be handled like resource links

carliros:
https://github.com/srid/emanote/issues/69

### DIFF
--- a/src/Emanote/Model/Link/Rel.hs
+++ b/src/Emanote/Model/Link/Rel.hs
@@ -80,6 +80,7 @@ unresolvedRelsTo r =
 parseUnresolvedRelTarget :: [(Text, Text)] -> Text -> Maybe UnresolvedRelTarget
 parseUnresolvedRelTarget attrs url = do
   guard $ not $ "://" `T.isInfixOf` url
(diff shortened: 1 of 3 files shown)


src/Emanote/Model/Link/Rel.hs
```diff
@@ -80,6 +80,7 @@ unresolvedRelsTo r =
 parseUnresolvedRelTarget :: [(Text, Text)] -> Text -> Maybe UnresolvedRelTarget
 parseUnresolvedRelTarget attrs url = do
   guard $ not $ "://" `T.isInfixOf` url
+  guard $ not $ "mailto:" `T.isInfixOf` url
```
srid, line 83: I wonder if we should replace both of this with a simple ```":" `T.isInfixOf` url``` ... because, we do want to ignore other protocols like ftp://, magnet:, etc.

Consequently, this would mean the note filenames cannot have `:` in them. I'll look to see if this is really necessary or can be avoided.

srid: Addressed in https://github.com/srid/emanote/commit/35ccd46ac6c83d3ead775eb9e3595def9db69c46

Issue #24, opened 2021-05-24
WikiLink Embedding - ![[..]]


srid:
Support for https://help.obsidian.md/How+to/Embed+files

- [x] Working prototype for .md and image files (#68)
- [x] Allow specifying note embed layout via Heist .tpl (#75)
- [x] ~~#88~~
- [x] #98 
- [ ] Other file types
  - [x] video
  - [x] audio
  - [x] pdf 
     - [x] #350
  - [x] Everything else? As code block. For eg., `foo.hs` gets syntax highlighted as haskell. https://github.com/srid/emanote/pull/719
    - [ ] Properties? ([file line nums](https://github.com/srid/neuron/issues/613), note snippet, ...)

vcavallo:
I propose adding an item to the checklist above: 

- [ ] Allow embedding a specific subset _block_ of a file. 

For instance, the `## Some Section` H2 heading and all text until next section. Obsidian's syntax is basically: `![[filename#Some Section]]`

Obsidian supports this and it would be great to have parity between Obsidian editing and Emanote's rendered results.

kukimik:
Embedding notes may create loops:

some_note.md:

```md
blah

![[other_note]]

blah
```

other_note.md:

```md
aaa

![[some_note]]

bbb
```

I've tried it with the live preview and couldn't open any of these notes in browser. Didn't have time to check memory/CPU usage etc., but perhaps the loops should be detected anyway?

srid:
@vcavallo Yes, we need to address it in a fundamental manner first: https://github.com/EmaApps/emanote/discussions/105

@kukimik Could you open a new issue?

srid:
I think we should also support source code embedding. For eg., `![[foo.hs]]` should include that as syntax highlighted code block. See also https://github.com/srid/neuron/issues/613 (not sure how that can be adopted for wiki links)

kukimik:
> @kukimik Could you open a new issue?

With the current version of emanote I can display the page. It looks like this:

![Peek 2022-10-11 22-10](https://user-images.githubusercontent.com/53443372/195189389-2710e2e3-8ec3-4be7-a89c-2a2bf4eb4262.gif)

It no longer loops indefinitely. (Or maybe it never did? I upgraded my hardware in the meantime; maybe I was just lacking resources - which however would mean the resource consumption was excessive.)

So I can open an issue, but I don't treat the current behavior as a bug/problem.

bolt12:
[...]

PR #75, opened 2021-06-20
Wikilink embedding, user customizable HTML template

srid:
#24


srid:
This PR uses a forked version of `heist` for https://github.com/snapframework/heist/issues/129 (and perhaps also more, going forward)

PR #76, opened 2021-06-20
Improve rendering of definition lists with a template

### LINKED ISSUE #32: Add <dl>, <dd> styling to markdown.tpl
The following Markdodwn

```markdown
Foo
: some text
Bar
: more text
```

generates a HTML definition list per [this extension](https://github.com/jgm/commonmark-hs/blob/master/commonmark-extensions/test/definition_lists.md) in use ... but it doesn't look like one (given Tailwind's reset). 

Allowing styling this list via the Heist template `pandoc.tpl` like we style, say, `BlockQuote`.

---

[This may be slightly harder, but it will make you familiarize with the Heist HTML template library]

Use `BlockQuote`'s implementation for inspiration,

https://github.com/srid/emanote/blob/78c9a2cf39d20fd90015e810b6d244cc270df49e/default/templates/components/pandoc.tpl#L12-L17

https://github.com/srid/emanote/blob/78c9a2cf39d20fd90015e810b6d244cc270df49e/src/Heist/Extra/Splices/Pandoc/Render.hs#L83-L85

Though, footnotes list is a better source, given that we need *iterate* over a list,

https://github.com/srid/emanote/blob/78c9a2cf39d20fd90015e810b6d244cc270df49e/default/templates/components/pandoc.tpl#L23-L37

https://github.com/srid/emanote/blob/78c9a2cf39d20fd90015e810b6d244cc270df49e/src/Heist/Extra/Splices/Pandoc/Footnotes.hs#L39-L49

carliros:
Resolves https://github.com/srid/emanote/issues/32

### DIFF
--- a/default/templates/components/pandoc.tpl
+++ b/default/templates/components/pandoc.tpl
@@ -15,6 +15,22 @@
       <blocks />
     </blockquote>
   </BlockQuote>
+  <DefinitionList>
+    <dl>
+      <definitionList:content>
+        <div class="px-4 py-5 sm:grid sm:grid-cols-${definitionList:grid-cols} sm:gap-4 sm:px-6">
+          <dt class="text-sm font-medium text-gray-500">
+            <definitionTerm:content />
+          </dt>
+          <definitionDescription:root>
+            <dd class="mt-1 text-sm text-gray-900 sm:mt-0">
+              <definitionDescription:content />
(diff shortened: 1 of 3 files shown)


srid:
Refactored a bit before merging to master, https://github.com/srid/emanote/commits/master

PR #92, opened 2021-06-28
Fix wiki link can have same title as alternative text

### LINKED ISSUE #77: Custom wiki-link text doesn't work if it is same as filename
1. Create foo.md and add `# Foo Title` to it.
2. In another note, link to it as `[[foo]]` and verify that it appears at "Foo Title".
3. Now use a custom text, as `[[foo|foo]]`, and notice the bug - of nothing changing.

Possible fix: compare with title instead of url in here: https://github.com/srid/emanote/blob/d4e973e8f821a2f921ad8497b425c03cf19dab94/src/Emanote/Pandoc/Filter/Url.hs#L79-L81

carliros:
Resolves https://github.com/srid/emanote/issues/77


srid:
Looks good. I've removed a import warning, and copy-edited the documentation text.

PR #100, opened 2021-07-01
Fix duplicate footnotes

### LINKED ISSUE #87: Duplicate footnotes when reusing the reference
If a footnote reference is used more than once, the corresponding footnote (at the bottom) will appear multiple times.

Repro,

```markdown
A paragraph[^1] with footnote.

Another location[^1] reusing that footnote.

[^1]: Footnote text.
```

When rendered, "Footnote text" will appear twice in the footnotes list at the bottom.

--

Bug appears to arise from here: https://github.com/srid/emanote/blob/852aad984900b726aec03ee4e35caa3139452a6a/src/Heist/Extra/Splices/Pandoc/Footnotes.hs#L17-L22

carliros:
Resolves https://github.com/srid/emanote/issues/87

### DIFF
--- a/emanote.cabal
+++ b/emanote.cabal
@@ -1,6 +1,6 @@
 cabal-version:      2.4
 name:               emanote
-version:            0.1.31.3
+version:            0.1.32.0
 license:            AGPL-3.0-only
 copyright:          2021 Sridhar Ratnakumar
 maintainer:         srid@srid.ca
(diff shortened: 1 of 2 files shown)


srid (approved):
As final one step, could you increment the version on the .cabal file @carliros? Going forward, let's do this for all PRs.

`List.nub` should work here without breaking the order of footnotes, as looking at [its implementation](https://hackage.haskell.org/package/base-4.15.0.0/docs/src/Data-OldList.html#nub) it appears to ensure that.

carliros:
@srid 
how do you decide which number to increment in the versioning?

srid:
@carliros Generally, I increment the third number (the first two are used for releases only) for any user-visible change. Otherwise the fourth (last) one. But don't worry about it, as long as you change the version in .cabal file some way, because then GitHub PR's interface gives me a way to directly edit it in the web browser (prior to merging).

Issue #99, opened 2021-07-01
List bullets are out of alignment in complex items


srid:
> I also wanted to point out that the bullet styling is a bit off sometimes. The text shows below the bullet point instead of beside it. 
> 
> ![image](https://user-images.githubusercontent.com/63365609/124153824-7a8c9700-dab2-11eb-9aee-e12d764bce82.png)
> 
> Here is an [example note:](https://n.xygn.us/C-Notes/Core/Pathology/Systemic/2.%20Inflammation)  
> [Source](https://gist.githubusercontent.com/xygn/10b6e9d75ef9368b40663d7b97bf03c9/raw/088195dc53c943070fb66c75b7d4302cebff31d9/Inflammation.md)

_Originally posted by @xygn in https://github.com/srid/emanote/discussions/97#discussioncomment-950194_

xygn:
[Generated HTML file](https://github.com/xygn/Notes/blob/gh-pages/C-Notes/Core/Pathology/Systemic/2.%20Inflammation.html)

[Source files](https://github.com/xygn/Notes)

srid:
Also happens when using tasklists and the list item overflows to next line,

![image](https://user-images.githubusercontent.com/3998/124207713-1cf94a00-dab4-11eb-889c-f29029c2fd84.png)

carliros:
Digging into this issue, looks like this is a pandoc formatting issue. I notice the issue is happening when there are two bullet lists separated by an empty line. When that happens, it is adding some paragraphs on list items.

For example:
```
Rendering lists

- item_1 list_1
- item_2 list_1

- item_1 list_2
```

The generated code is this
```
Para [Str "Rendering",Space,Str "lists"]
BulletList 
  [ [Para [Str "item_1",Space,Str "list_1"]]
  , [Para [Str "item_2",Space,Str "list_1"]]
  , [Para [Str "item_1",Space,Str "list_2"]]]
```

Those paragraphs are causing the alignment issue.

But, if we add some text to make sure that second list is a diffirent block, then it just works fine. Like this:
```
Rendering lists

- item_1 list_1
- item_2 list_1

Para
- item_1 list_2
```

This would end up as
```
Para [Str "Rendering",Space,Str "lists"]
BulletList 
  [ [Plain [Str "item_1",Space,Str "list_1"]]
  , [Plain [Str "item_2",Space,Str "list_1"]]]
Para [Str "Para"]
BulletList 
[...]

PR #103, opened 2021-07-05
Issue 99: Replace Para with Plain in List structures

carliros:
https://github.com/srid/emanote/issues/99

### DIFF
--- a/emanote.cabal
+++ b/emanote.cabal
@@ -1,6 +1,6 @@
 cabal-version:      2.4
 name:               emanote
-version:            0.1.37.2
+version:            0.1.38.1
 license:            AGPL-3.0-only
 copyright:          2021 Sridhar Ratnakumar
 maintainer:         srid@srid.ca
--- a/src/Heist/Extra/Splices/Pandoc/Render.hs
+++ b/src/Heist/Extra/Splices/Pandoc/Render.hs
@@ -86,12 +86,14 @@ rpBlock' ctx@RenderCtx {..} b = case b of
   B.BlockQuote bs ->
     withTplTag ctx "BlockQuote" ("blocks" ## rpBlock ctx `foldMapM` bs) $
       one . X.Element "blockquote" mempty <$> foldMapM (rpBlock ctx) bs
-  B.OrderedList _ bss ->
+  B.OrderedList _ bss' -> do
+    let bss = handleSpacing bss'
     withTplTag ctx "OrderedList" (pandocListSplices "OrderedList" bss) $ do
       fmap (one . X.Element "ol" (rpAttr $ bAttr b)) $
         flip foldMapM bss $
           fmap (one . X.Element "li" mempty) . foldMapM (rpBlock ctx)
-  B.BulletList bss ->
+  B.BulletList bss' -> do
+    let bss = handleSpacing bss'
     withTplTag ctx "BulletList" (pandocListSplices "BulletList" bss) $ do
       fmap (one . X.Element "ul" (rpAttr $ bAttr b)) $
         flip foldMapM bss $
@@ -184,6 +186,13 @@ rpBlock' ctx@RenderCtx {..} b = case b of
         itemsSplices :: [B.Block] -> H.Splices (HI.Splice n)
         itemsSplices bs = do
           (tagPrefix <> ":Item") ## foldMapM (rpBlock ctx) bs
+
+handleSpacing :: [[B.Block]] -> [[B.Block]]
+handleSpacing = map (map paraToPlain)
+
+paraToPlain :: B.Block -> B.Block
+paraToPlain (B.Para xs) = B.Plain xs
+paraToPlain x = x
 
 headerTag :: HasCallStack => Int -> Text
 headerTag n =
--- a/src/Heist/Extra/Splices/Pandoc/Render.hs
+++ b/src/Heist/Extra/Splices/Pandoc/Render.hs
@@ -87,13 +87,13 @@ rpBlock' ctx@RenderCtx {..} b = case b of
     withTplTag ctx "BlockQuote" ("blocks" ## rpBlock ctx `foldMapM` bs) $
       one . X.Element "blockquote" mempty <$> foldMapM (rpBlock ctx) bs
   B.OrderedList _ bss' -> do
-    let bss = handleSpacing bss'
+    let bss = convertFirstParaToPlain bss'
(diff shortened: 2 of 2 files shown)


srid (commented):
Wouldn't this break the scenario where a list item has multiple paragraphs? 

```markdown
- One paragraph
- Second item with two paragraphs. This is first para

  Second paragraph goes here. You can even have other block elements, like sub-lists.
- Third item
```

Also let's put the new functions in the `where` clause of the only function where they get used. As well favour `fmap` to `map` for consistency. What does `handleSpacing` do (semantically)? This function needs a code comment, because it seems to have a higher meaning than a mere double fmap (if it is the latter, we don't need a separate function for it).

srid:
By the way, you can use `Resolves #99` in PR description; then GitHub will automatically close the issue when this PR gets merged. https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

carliros:
@srid 
Good catch on the multiple Para's. I am applying the Para2Plain only to the first block that matches on each item list. So that next block elements are not touched.

I made some refactorings as well, thx for the feedback.

srid:
Wait, I'm now confused as to what this is intended to achieve. Could you add demo to `markdown.md` illustrating what this change does, visually?

What does it look like for each of these?

```markdown
Simple list with one paragraph in each item

- Foo
- Bar
- Qux
```

```markdown
Second item is multi-paragraph

- Foo
- Bar

  Second paragraph
- Qux
```

Per `applyFirst ((B.Para xs) : bss) = B.Plain xs : bss` it seems that we are eliminating the wrapping `<p>` for `Bar` (breaking the semantic of "two paragraph" item, because now there is just one paragraph). Why do we do this? If it is just to avoid the margins, shouldn't that be done in CSS? Or maybe I'm missing something.

srid:
> I just see commonmark-pandoc does the same as your suggestion!
>
> https://github.com/jgm/commonmark-hs/blob/344d380edc1f44ab476d176050263b92d9bfd411/commonmark-pandoc/src/Commonmark/Pandoc.hs#L160
>
> _Originally posted by @carliros in https://github.com/srid/emanote/issues/99#issuecomment-874258042_

Looking at what commonmark does, https://github.com/jgm/commonmark-hs/blob/344d380edc1f44ab476d176050263b92d9bfd411/commonmark-pandoc/src/Commonmark/Pandoc.hs#L156-L158

They convert *all* para to plain if `TightList` is used, as opposed to `LooseList`. The logic to pick this is here, https://github.com/jgm/commonmark-hs/blob/7958e5b40093fc0cd4ed3adc306befb9352344e2/commonmark/src/Commonmark/Blocks.hs

Since Emanote already uses commonmark, this code should already be in effect, no? Maybe my original suggestion is wrong. Hard to assess this without "before" and "after" demos.

carliros:
@srid 
looks like it is not in effect as there is Para's in BulletList.

Here is how it looks without the changes:
Screen Shot 2021-07-07 at 13 34 19

And here is with my changes:
Screen Shot 2021-07-07 at 13 34 49

srid:
Looks like commonmark is using `TightList` in the first case (because **every** item is a one-para item), but `LooseList` if even one of the item has more than one para. And only `TightList` does does the para->plain conversion (the first list in your example, but not second).

There are a couple of reasons why we need an alternative solution here:

1. Notice the lack of margin space between "Bar" and "Second paragraph" in the "after" image. This happens because "Bar" is no longer a paragraph (with this PR's change), and so there is no below margin (`mb-3`). But visually we expect margins to be present between two paragraphs.
2. This issue happens not just with multi-para but with other block elements as well, such as [tasklists](https://github.com/srid/emanote/issues/99#issuecomment-872650108).

I thought that the solution I proposed below should address both these issues, thus resolving #99 in its entirety:

> one solution here is to stop using *native* list bullets (ie, the `list-disc` class) entirely, and do our own bullets in CSS using flexbox (we already do this for definition lists).
> 
> _Originally posted by @srid in https://github.com/srid/emanote/issues/99#issuecomment-873241604_

However, this would work only for `BulletList` but not for `OrderedList`. So it is not ideal either.

Regardless, #99 can only be fixed by fixing both (1) and (2).

srid:
With the following patch,

```diff
diff --git a/src/Heist/Extra/Splices/Pandoc/Render.hs b/src/Heist/Extra/Splices/Pandoc/Render.hs
index c1b9a66..5edab98 100644
--- a/src/Heist/Extra/Splices/Pandoc/Render.hs
+++ b/src/Heist/Extra/Splices/Pandoc/Render.hs
@@ -91,7 +91,7 @@ rpBlock' ctx@RenderCtx {..} b = case b of
     withTplTag ctx "BulletList" (pandocListSplices "BulletList" bss) $ do
       fmap (one . X.Element "ul" (rpAttr $ bAttr b)) $
         flip foldMapM bss $
-          fmap (one . X.Element "li" mempty) . foldMapM (rpBlock ctx)
+          fmap (one . X.Element "li" mempty) . foldMapM (rpBlock ctx) . singletonListItem
   B.DefinitionList defs ->
     withTplTag ctx "DefinitionList" (definitionListSplices defs) $
       fmap (one . X.Element "dl" mempty) $
@@ -179,7 +179,14 @@ rpBlock' ctx@RenderCtx {..} b = case b of
       where
         itemsSplices :: [B.Block] -> H.Splices (HI.Splice n)
         itemsSplices bs = do
-          (tagPrefix <> ":Item") ## foldMapM (rpBlock ctx) bs
+          (tagPrefix <> ":Item") ## foldMapM (rpBlock ctx) $ singletonListItem bs
+
+    -- Refine commonmark's LooseList result. Unpara all items with single
+    -- paragraph, except those with multiple paragraphs.
+    singletonListItem :: [B.Block] -> [B.Block]
+    singletonListItem = \case
+      [B.Para is] -> [B.Plain is]
+      x -> x
```

I get the following results in place of https://note.ema.srid.ca/demo/markdown#lists ,

image

Basically it narrows the bug to just list items with more than one paragraph.  But if you add a CSS class `inline-block` to the *first* paragraph, the problem vanishes. But then the second paragraph remains deindented, which is a problem.

srid:
Turns out the culprit is [`list-inside`](https://tailwindcss.com/docs/list-style-position#usage). I don't understand why exactly, but removing it fixed both the issues (and obviates this PR; so closing).

https://github.com/srid/emanote/commit/74d593a37780bb353607cd7f335061abf3cfe547

image

image

PR #108, opened 2021-07-16
Add support for inline image embeding

### LINKED ISSUE #98: Allow inline embedding images
The following works,

```markdown
Para 1

![[Foo.jpeg]]

Para 2
```

But the following does not,

```markdown
A paragraph that contains an image: ![[Foo.jpeg]]
```

See https://github.com/srid/emanote/discussions/97 for background

carliros:
Resolves https://github.com/srid/emanote/issues/98

### DIFF
--- a/default/templates/filters/embed-image.tpl
+++ b/default/templates/filters/embed-image.tpl
@@ -1,1 +1,1 @@
-<img class="mb-3" src="${ema:url}" alt="${ema:alt}" />
+<img class="inline mb-3" src="${ema:url}" alt="${ema:alt}" style="width: inherit;" />
--- a/docs/demo/embed.md
+++ b/docs/demo/embed.md
@@ -28,6 +28,8 @@ Embedding image files as, say, `![[disaster-girl.jpg]]` is equivalent to `![](pa
 
 ![[disaster-girl.jpg]]
 
+It is also posible to add images inline (example, here's the site favicon: [![[favicon.svg]]]{.w-6}) say in the middle of a paragraph.
+
 ### Videos
 
 The following is the result of using `![[death-note.mp4]]`.
--- a/src/Emanote/Pandoc/Renderer/Embed.hs
+++ b/src/Emanote/Pandoc/Renderer/Embed.hs
@@ -13,6 +13,10 @@ import qualified Emanote.Model.Title as Tit
 import Emanote.Pandoc.BuiltinFilters (prepareNoteDoc, preparePandoc)
 import qualified Emanote.Pandoc.Markdown.Syntax.WikiLink as WL
 import Emanote.Pandoc.Renderer (PandocBlockRenderer, PandocInlineRenderer)
+import Emanote.Pandoc.Renderer.BrokenLink
+  ( BrokenLink (BrokenLink_Block, BrokenLink_Inline),
+    renderBrokenLink,
+  )
 import qualified Emanote.Pandoc.Renderer.Url as Url
 import qualified Emanote.Route as R
 import qualified Emanote.Route.SiteRoute as SR
@@ -102,24 +106,3 @@ videoExts =
     ".webm",
     ".ogv"
   ]
-
--- | Like Pandoc's `Link` node, but for denoting "broken" links.
-data BrokenLink
-  = BrokenLink_Block B.Attr [B.Inline] (Text, Text)
-  | BrokenLink_Inline B.Attr [B.Inline] (Text, Text)
-  deriving (Eq, Show)
-
-renderBrokenLink :: Monad n => Model -> HP.RenderCtx n -> Text -> BrokenLink -> HI.Splice n
-renderBrokenLink model ctx err = \case
-  BrokenLink_Block attr is x ->
-    HP.rpBlock ctx $
-      B.Div (Url.brokenLinkAttr err) $
-        one . B.Para . one $
-          B.Link attr (fixIs (fst x) is) x
-  BrokenLink_Inline attr is x ->
-    HP.rpInline ctx $
-      B.Span (Url.brokenLinkAttr err) $
-        one $ B.Link attr (fixIs (fst x) is) x
-  where
-    fixIs url is =
-      Url.nonEmptyLinkInlines model url Nothing is
--- a/src/Emanote/Pandoc/Renderer/Embed.hs
+++ b/src/Emanote/Pandoc/Renderer/Embed.hs
@@ -24,7 +24,6 @@ import qualified Heist as H
 import qualified Heist.Extra as HE
 import Heist.Extra.Splices.Pandoc (pandocSplice)
 import qualified Heist.Extra.Splices.Pandoc as HP
-import Heist.Extra.Splices.Pandoc.Ctx (ctxSansCustomSplicing)
 import qualified Heist.Interpreted as HI
 import qualified Text.Pandoc.Definition as B
 
@@ -38,7 +37,7 @@ embedBlockWikiLinkResolvingSplice _emaAction model _nf ctx _ blk =
       case Url.resolveWikiLinkMustExist model wl of
         Left err -> do
           let brokenLink = BrokenLink_Block attr is (url, tit)
-          pure $ renderBrokenLink (ctxSansCustomSplicing ctx) err brokenLink
+          pure $ renderBrokenLink ctx err brokenLink
         Right res -> do
           embedBlockSiteRoute model ctx res
     _ ->
@@ -52,7 +51,7 @@ embedInlineWikiLinkResolvingSplice _emaAction model _nf ctx _ inl = case inl of
     case Url.resolveWikiLinkMustExist model wl of
       Left err -> do
         let brokenLink = BrokenLink_Inline attr is (url, tit)
-        pure $ renderBrokenLink (ctxSansCustomSplicing ctx) err brokenLink
+        pure $ renderBrokenLink ctx err brokenLink
       Right res -> do
         embedInlineSiteRoute wl res
   _ -> Nothing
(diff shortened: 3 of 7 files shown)


carliros:
Here is a screenshot: 

Screen Shot 2021-07-16 at 14 11 36

default/templates/filters/embed-image.tpl
```diff
@@ -1 +1 @@
-<img class="mb-3" src="${ema:url}" alt="${ema:alt}" />
\ No newline at end of file
+<img class="mb-3" src="${ema:url}" alt="${ema:alt}" style="display: inline;" />
```
srid: You can replace this with Tailwind's [`inline`](https://tailwindcss.com/docs/display#inline) class. 

More generally we should use Tailwind's classes wherever applicable, and resort to `style` attribute only when absolutely necessary. Tailwind is a feature of the default templates (you can use them index.yaml for instance).

src/Emanote/Pandoc/Renderer/Embed.hs
```diff
@@ -43,14 +43,41 @@ embedWikiLinkResolvingSplice emaAction model nf (ctxSansCustomSplicing -> ctx) _
         B.Div (Url.brokenLinkAttr err) $
           one block
 
-embedSiteRoute :: Monad n => Ema.CLI.Action -> Model -> NoteRenderers n -> HP.RenderCtx n -> WL.WikiLink -> Either MN.Note SF.StaticFile -> Maybe (HI.Splice n)
-embedSiteRoute emaAction model nf RenderCtx {..} wl = \case
+embedInlineWikiLinkResolvingSplice ::
+  Monad n => PandocInlineRenderer n x
+embedInlineWikiLinkResolvingSplice emaAction model nf (ctxSansCustomSplicing -> ctx) _ inl =
+  case inl of
+    B.Link attr@(_id, _class, otherAttrs) is (url, tit) -> do
```
srid: You can use `LambdaCase` here instead of using `inl`?

src/Emanote/Pandoc/Renderer/Embed.hs
```diff
@@ -43,14 +43,41 @@ embedWikiLinkResolvingSplice emaAction model nf (ctxSansCustomSplicing -> ctx) _
...
       "ema:note:pandoc"
         ## noteSpliceWith nf classMap emaAction model note
+  Right _ -> Nothing
+  where
+    runEmbedTemplate name splices = do
+      tpl <- HE.lookupHtmlTemplateMust $ "/templates/filters/embed-" <> name
+      HE.runCustomTemplate tpl splices
+
+embedInlineSiteRoute :: Monad n => Ema.CLI.Action -> Model -> NoteRenderers n -> HP.RenderCtx n -> WL.WikiLink -> Either MN.Note SF.StaticFile -> Maybe (HI.Splice n)
+embedInlineSiteRoute _ _ _ _ wl = \case
```
srid: As a general Haskell coding guideline, if an argument is never used in a function, it should be removed from its signature. These 4 arguments should be removed. 

Why? The functional signature should be as simple as possible for understanding. The signature in part acts as its 'documentation', in that - reading the signature more or less tells you what this function does.

src/Emanote/Pandoc/Renderer/Embed.hs
```diff
@@ -43,14 +43,41 @@ embedWikiLinkResolvingSplice emaAction model nf (ctxSansCustomSplicing -> ctx) _
...
+  Left _ -> Nothing
```
srid: Could you add comment stating *why* this case is `Nothing`?

EDIT: see other comment.

src/Emanote/View/Template.hs
```diff
@@ -93,8 +93,10 @@ renderLmlHtml emaAction model note = do
       pageTitle = M.modelLookupTitle r model
       noteRenderers =
         NoteRenderers
-          [PF.urlResolvingSplice]
-          [ PF.embedWikiLinkResolvingSplice,
+          [ PF.embedInlineWikiLinkResolvingSplice,
+            PF.urlResolvingSplice
```
srid: Is there a reason for changing the order of `urlResolvingSplice` (it is no longer the first spice in this PR)?

docs/demo/embed.md
```diff
@@ -28,6 +28,8 @@ Embedding image files as, say, `![[disaster-girl.jpg]]` is equivalent to `![](pa
 
 ![[disaster-girl.jpg]]
 
+It is also posible to add ![[ema-small.png]] as inline paragraphs.
```
srid: Do we really need to add a new image file to the repository? Can't we use `favicon.svg`? Shouldn't something like this work?

```markdown
It is also posible to add images inline (example, here's the 
site favicon: [![[favicon.svg]]]{.w-6}) say in the middle of a paragraph.
```

src/Emanote/Pandoc/Renderer/Embed.hs
```diff
@@ -43,14 +43,41 @@ embedWikiLinkResolvingSplice emaAction model nf (ctxSansCustomSplicing -> ctx) _
...
+
+embedBlockSiteRoute :: Monad n => Ema.CLI.Action -> Model -> NoteRenderers n -> HP.RenderCtx n -> WL.WikiLink -> Either MN.Note SF.StaticFile -> Maybe (HI.Splice n)
+embedBlockSiteRoute emaAction model nf RenderCtx {..} wl = \case
   Left note -> do
     pure . runEmbedTemplate "note" $ do
       "ema:note:title" ## Tit.titleSplice (preparePandoc model) (MN._noteTitle note)
       "ema:note:url" ## HI.textSplice (SR.siteRouteUrl model $ SR.lmlSiteRoute $ note ^. MN.noteRoute)
       "ema:note:pandoc"
         ## noteSpliceWith nf classMap emaAction model note
+  Right _ -> Nothing
```
srid: Could you add comment stating why this case is Nothing?

EDIT: see other comment.

src/Emanote/Pandoc/Renderer/Embed.hs
```diff
@@ -43,14 +43,41 @@ embedWikiLinkResolvingSplice emaAction model nf (ctxSansCustomSplicing -> ctx) _
...
+  where
+    runEmbedTemplate name splices = do
```
srid: It looks like `runEmbedTemplate ` is used in the other function as well? If so it should be made a top-level, an reused without duplication.

src/Emanote/Pandoc/Renderer/Embed.hs
```diff
@@ -43,14 +43,41 @@ embedWikiLinkResolvingSplice emaAction model nf (ctxSansCustomSplicing -> ctx) _
...
+  Monad n => PandocInlineRenderer n x
+embedInlineWikiLinkResolvingSplice emaAction model nf (ctxSansCustomSplicing -> ctx) _ inl =
+  case inl of
+    B.Link attr@(_id, _class, otherAttrs) is (url, tit) -> do
+      Rel.URTWikiLink (WL.WikiLinkEmbed, wl) <- Rel.parseUnresolvedRelTarget (otherAttrs <> one ("title", tit)) url
+      case Url.resolveWikiLinkMustExist model wl of
+        Left err ->
+          pure $ brokenLinkSpanWrapper err $ B.Link attr (Url.nonEmptyLinkInlines model url Nothing is) (url, tit)
+        Right res -> do
+          embedInlineSiteRoute emaAction model nf ctx wl res
```
srid: Okay, as a follow-up to two other comments on `Nothing` further below, I think it would be better to refactor this into two functions. One that takes `noteRoute` as argument and returns a block embedding splice (call it `embedResourceRoute`); another that takes `staticFileRoute` as argument and returns an inline embedding splice (call it `embedStaticFileRoute`). 

Then you can use it here as,

```haskell
either (embedResourceNote emaAction model nf ctx wl) (const Nothing) res
```

and in the inline splice,

```haskell
either (const Nothing) (embedStaticFileRoute emaAction model nf ctx wl) res
```

srid (changes requested):
Thanks! Added my feedback.

srid:
Also .cabal file version should be incremented.

carliros:
Super cool, thanks for the feedback, I will address each of them. I am learning a lot on participating on this project.

srid:
@carliros #109 refactored the renderer types; you will want to merge from `master` to resolve the conflicts. Let me know if the conflict resolution is confusing, I can help.

EDIT: It might be better to re-do your changes on top of current `master`

src/Emanote/View/Template.hs
carliros: This was because having `urlResolvingSplice` first does not recognize the inline element (Link) as image but as File embed but not rendered as image. Putting first `PF.embedInlineWikiLinkResolvingSplice` fixed that.

docs/demo/embed.md
carliros: Good to know that we can use that image as well.

src/Emanote/View/Template.hs
srid: In that case, we should add a code comment saying exactly that (for posterity).

src/Emanote/Pandoc/Renderer/Embed.hs
carliros: being addressed with the refactoring suggested in other comment.

src/Emanote/Pandoc/Renderer/Embed.hs
carliros: being addressed with the refactoring suggested in other comment.

carliros:
@srid I am done with the refactoring. Please take a look.

carliros:
One think might need some extra explanation is the add of `width: inherit` in
```
<img class="inline mb-3" src="${ema:url}" alt="${ema:alt}" style="width: inherit;" />
```

That is needed in order to respect the father width with the child element. Without that, the image is rendered in full width.

srid (approved):

srid:
Just refactored a few things, mainly the "broken link" handling/rendering was spread across Url.hs and Embed.hs; I've consolidated it into one place, so in future it will be simpler to reason about this concept from multiple places. As a meta/pedagogic note, it is quite common to do a "refactor postmortem" on feature PRs like this, so as to make sure that avoidable complexity can be kept at bay (as the code size grows).

EDIT: btw, this refactor also uncovered a hang bug, as fixed in the first commit.

PR #112, opened 2021-07-26
Add menu icon to show/hide sidebar

### LINKED ISSUE #13: Toggle-able sidebar
Change the HTML template to improve navigation on mobile. Around here,

https://github.com/srid/emanote/blob/78c9a2cf39d20fd90015e810b6d244cc270df49e/default/templates/layouts/book.tpl#L25-L28

- [ ] In mobile view (where sidebar currently remains always hidden), display the burger menu on top, clicking which toggles the sidebar visibility

Inspiration / example: https://web.dev/learn/css/flexbox/

carliros:
Resolves https://github.com/srid/emanote/issues/13

### DIFF
--- a/default/templates/layouts/book.tpl
+++ b/default/templates/layouts/book.tpl
@@ -4,7 +4,7 @@
 
       <apply template="components/breadcrumbs" />
 
-      <div class="flex flex-nowrap bg-gray-50 md:mt-8 md:shadow-2xl md:mb-8">
+      <div id="container" class="flex flex-nowrap bg-gray-50 md:mt-8 md:shadow-2xl md:mb-8">
         <!-- Sidebar column -->
         <nav id="sidebar"
           class="flex-shrink hidden leading-relaxed md:block md:sticky md:top-0 md:h-full md:w-48 xl:w-64">
@@ -70,6 +70,13 @@
           </main>
         </div>
       </div>
+      <style type="text/css">
+        @media (min-width: 768px) {
+          #container {
+            display: flex;
+          }
+        }
+      </style>
       <apply template="components/footer" />
     </div>
   </bind>
--- a/default/templates/layouts/book.tpl
+++ b/default/templates/layouts/book.tpl
@@ -4,7 +4,8 @@
 
       <apply template="components/breadcrumbs" />
 
-      <div id="container" class="flex flex-nowrap bg-gray-50 md:mt-8 md:shadow-2xl md:mb-8">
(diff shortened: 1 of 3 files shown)


carliros:
@srid I have added a button to show/hide sidebar section. I see two issues:
1. On index page, no breadcrumb is show, then no menu icon as well
2. Changing from small screen (with menu sidebar is shown) to a bigger screen (where menu sidebar icon is not shown), page does not render well. I think this is not a common case, but it is an issue.

carliros:
Some screenshots:

Screen Shot 2021-07-26 at 14 31 40

Screen Shot 2021-07-26 at 14 31 55

srid:
@carliros I'll address (1). As for (2), try a no-JS approach: https://codepen.io/mutedblues/pen/MmPNPG

srid:
If JS is necessary, then: https://codepen.io/ttntm/pen/dqaNPp

srid:
@carliros I'll address (1). As for (2),  try https://ttntm.me/blog/tailwind-responsive-menu/

(There is also a [pure-css solution](https://codepen.io/mutedblues/pen/MmPNPG)).

srid:
> On index page, no breadcrumb is show, then no menu icon as well

Addressed in 9f7218b

carliros:
Hi @srid, I fixed (2) issue. So this is ready to go. Please take a look.

default/templates/layouts/book.tpl
```diff
@@ -70,6 +70,13 @@
           </main>
         </div>
       </div>
+      <style type="text/css">
+        @media (min-width: 768px) {
+          #container {
+            display: flex;
+          }
```
srid: Hmm, not sure why this is needed. `flex` is already in the class list of `id=container` div further above?

carliros: that is needed to resolve issue (2). The idea is that when we are on small screen and we click on menu button, then "flex" class is removed from "container" tag. Then, if we change the size to a bigger one so that menu button is hidden, there were no way to put back the "flex" class there were removed as part of clicking on menu button. So, that CSS forces the tag "container" to put back flex mode for that tag and it is showing properly... that is how I am fixing it.

srid: You can solve this by switching between `flex-col` and `flex-row` when viewport changes (Tailwind's `md:` thing does it - see https://tailwindcss.com/docs/responsive-design ) ... just pushed a commit.

PR #149, opened 2021-09-24
Remove erroneous -L flag, clarify README

spanners:
As can be seen from running emanote, there `-L` is an invalid option. This patch removes it from the README.

```
$  mkdir /tmp/output && emanote -L /home/simon/zettel gen /tmp/output                                                                                                                      
Invalid option `-L'

Usage: emanote [--version] [--layers LAYERS] [COMMAND]
  Emanote - A spiritual successor to Neuron
  ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,9 +23,9 @@ nix-env -if https://github.com/srid/emanote/archive/refs/heads/master.tar.gz
 cd /path/to/notebook
 PORT=8001 emanote
 
-# Generate static files
+# Or, to generate static files
 mkdir /tmp/output
-emanote -L /path/to/notebook gen /tmp/output
+emanote /path/to/notebook gen /tmp/output
 ```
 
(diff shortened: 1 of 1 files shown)


srid:
@spanners Which version of `emanote` are you using? `-L` is a valid flag since https://github.com/srid/emanote/commit/8c4f36f69ec2cde2693263bfc5cf7e3af495b7f8 (about 2 days ago) and it is an alias to `--layers`.

(The layers concept does more than `cd`'ing to the directory, as it allows you to merge multiple notebooks. It is not fully documented yet ... but you can see it being used [here](https://github.com/srid/emanote-template/blob/99ddf93ca3cea47f2fd44b04c93c8c57d39d63ab/.github/workflows/publish.yaml#L15) to override settings for GitHub Pages deployment).

spanners:
ok, in which case I'll close the issue. thanks :)

PR #155, opened 2021-09-25
nixpkgs -> haskell.nix

srid:
- [x] `bin/run` works
- [ ] `nix-env -if .` works
- [ ] doesn't build too much
    - builds nix
- [ ] CI iohk cache
- [ ] cross-compile musl?

### DIFF
--- a/bin/run
+++ b/bin/run
@@ -6,4 +6,4 @@ export PORT=9010
 # point, with the specified args.
 #
 # If you change ./.ghcid, ghcid will automatically reload.
-exec nix develop -c ghcid
+exec nix develop -j auto -c ghcid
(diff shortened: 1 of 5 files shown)


srid:
> - [ ] `nix-env -if .` works

For some reason, this gets stuck at:

```
copying path '/nix/store/hd4qr94hhq8d1xrhqq3lls1zf6qrs94y-01-index.tar.gz-at-2021-09-22T000000Z' from 'https://hydra.iohk.io'...
```

EDIT: resumed after a few re-runs.

srid:
> - [ ] doesn't build too much

Nope, `nix build` (but not `nix develop -c ghcid`) builds aeson and many other packages (if not ghc itself).

(It even builds `lvar`, which doesn't transitively depend on `heist` that we override)

Issue #161, opened 2021-10-03
Expose JSON dump of notebook metadata


srid:
Add `emanote export` so as to be able to export a JSON dump of the graph metadata (including backlink contexts, if requested) of the notebook.

(This would be analogous to `neuron query -g`)

Consider implementing `emanote gen --with-export` so that the dump is generated in the static site as well.

edrex:
i'm plotting a command-line wiki refactoring tool (ops like rename/move with link rewriting, subgraph-based wiki splitting, document format linting), and access to emanote's link graph would be interesting.

srid:
For those looking to implement this, my initial thought is that we *essentially* only need to convert these three fields in the model to JSON:

https://github.com/srid/emanote/blob/149774c7d0bce92f2a55d32e9834ab5812fc9cb9/src/Emanote/Model/Type.hs#L46-L48

(The first provides the notes themselves, including the parsed Pandoc AST; the second provides the relationships between notes; the third provides the YAML metadata).

In addition to adding a CLI command that invokes runs `Source.emanate` in an one-off manner (to get the initial model).

srid:
@edrex The linked PR (in draft form) #174 implements some version of the export feature. Could you try it out for your use cases and provide feedback?

You can clone that branch and run `nix-env -if ./default.nix` (or `nix-build -o ./output`) to use that custom Emanote. Then, either in live server or generated static site, access the URL path `/-/export.json` to get the graph in JSON. It should look like this (with [the formatter](https://chrome.google.com/webstore/detail/json-formatter/bcjindcccaagfpapjjmafapmmgkkhgoa/related?hl=en)):

![image](https://user-images.githubusercontent.com/3998/136713137-241b126f-1ccc-4488-a7e3-4b3f7431932b.png)

(This includes only the relations, and I'll shortly expose the note metadata as well).

edrex:
Tested briefly on my big messy wiki. It's fast, presumably exposing an existing in-memory structure.

Impressions:

- `tag: RTMissing` edges lack contents field, which would be handy for enumerating missing link targets:

```
> curl -s http://localhost:8000/-/export.json|jq '[.[] | select(.[1].tag == "RRTMissing")][-1]' 
["jq.md", {"tag": "RRTMissing"}]
```

- `/-/export.json` is present in the generated output as well, making it available to eg, build a force-directed graph client-side, which seems like a fun possibility

Reading and understanding Haskell is a very slow process for me rn so I can't comment on the code.

[...]

PR #174, opened 2021-10-10
Dedicate a route to get the JSON of graph data

srid:
For #161

- [x] MVP
- [x] Normalize paths to use slug always (helps with client-side scripting)
- [x] Add JSON version
- [ ] Include non-note files (static files) in vertex list
- [ ] Include per-route .yaml meta?
 - [ ] Make json keys and such concise enough, to reduce output size on large zks

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -24,7 +24,7 @@ jobs:
           signingKey: "${{ secrets.CACHIX_SIGNING_KEY }}"
           # Only needed for private caches
           authToken: "${{ secrets.CACHIX_AUTH_TOKEN }}"
-      - name: Build docker img
+      - name: Build as docker img
         id: build
         run: |
           nix-build -j auto docker.nix --argstr tag latest -o docker-img
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+  pull_request:
 jobs:
 
   build:
@@ -32,6 +33,7 @@ jobs:
           echo "::set-output name=docker-img::${DOCKERIMG}"
       - name: Upload docker image tgz
         uses: actions/upload-artifact@v2
+        if: github.ref == 'refs/heads/master'
         with:
           name: docker-img
           retention-days: 1
@@ -41,6 +43,7 @@ jobs:
   docker:
     needs: build
     runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
     env:
       DOCKERTAG: latest
     steps:
@@ -58,6 +61,7 @@ jobs:
   website:
     needs: build
     runs-on: ubuntu-latest
+    if: github.ref == 'refs/heads/master'
     steps:
       - uses: actions/checkout@v2
       - name: Download docker image
--- a/docs/demo/file-links.md
+++ b/docs/demo/file-links.md
@@ -7,6 +7,7 @@ For example,
 - Here is a link to some text file: [[Sample.txt]]
 - You can also specify the full, or subset of, the path: [[demo/Sample.txt]]
 - Of course, a custom link text may also be specified: [[Sample.txt|Some text file]]
+- All of this is equivalent to [normal linking](demo/Sample.txt).
 
 In the live server, links to static files will open in new browser tab.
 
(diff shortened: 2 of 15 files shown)


edrex:
Tested some on my big messy wiki. I have a few links of the form `[](target)` (empty text); these produce an edge with `type(unresolvedRelTarget.contents) == 'string'`) which makes mapping over edges awkward:

```
~/src/github.com/srid/emanote export-route* ❯ curl -s http://localhost:8000/-/export/graph.json|jq '.edges | to_entries[] | .value[] as $v | {key: .key, value: $v} | select(.value.unresolvedRelTarget.contents | type == "string")'
{
  "key": "Bar",
  "value": {
    "resolvedRelTarget": {
      "tag": "RRTMissing"
    },
    "unresolvedRelTarget": {
      "tag": "URTResource",
      "contents": "Created"
    }
  }
}
...
```

This little gem also has the same issue, presumably because the formatted text and whitespace are stripped: `[ *foo* **bar**](hi)`.

**Update** to be clear, i think the expected behavior would be for the `contents` field to have a value like:

```
[
  "MarkdownLink", // or some such
  "[](foo)"
]
```

**Update 2** 

- It looks like the contents field is a string for all local/relative markdown-style links
- Not directly-related to this issue, but I see that relative markdown links like `[foo](../foo.md)` and [baz](./baz.md) are erroneously tagged as `RRTMissing`. This might be part of a larger conversation about which of the various possible local/relative reference formats to support in both wiki links and markdown links, so I'll defer FTM

edrex:
I really, really like the idea of decorating standard wiki links as `#[[Parent]]` and `[[Child]]#`; it seems to allow completely replacing tags with pages (i've called them "emboddied tags" in other formats like TiddlyWiki). It would be cool if those parent/child links were indicated in the graph export, although I see that it's tracked in a different model so maybe that will have to wait. That model also constructs implicit parent links like `Foo/Bar.md => Foo.md`.

If the export contained these relns, it would be possible to visualize the heirachical graph structure, then overlay a messy force directed graph of the rest of the links, which I think could be interesting for reasoning about possible heirarchies. But again, it seems like that might make sense in another iteration.

srid:
I removed including raw link syntax in JSON to avoid this confusion. So instead of `[[Foo bar]]` it now shows `Foo bar`. Thus, to be consistent, we would use regular Markdown links (eg: `foo-bar.md`). 

An empty link such as `[](empty-link.md)` produces this JSON:

```json
{
  "resolvedRelTarget": {
    "tag": "RRTMissing"
  },
  "unresolvedRelTarget": {
    "tag": "URTResource",
    "contents": "empty-link.md"
  }
}
```

Okay, that looks fine to me? The JSON is supposed to give you the (broken) URL, not the link text per se.

"contents" is simply the default key used by Haskell's JSON encoder for denoting the arguments of a sum type's construtor. The "tag" refers to the constructor itself. Here's the sum type in question:

https://github.com/srid/emanote/blob/069dcc746ea9b55cd149befee0f9048ae8fc6f1a/src/Emanote/Model/Link/Rel.hs#L39-L47

which is the "unresolved" version of the resolved one:

https://github.com/srid/emanote/blob/069dcc746ea9b55cd149befee0f9048ae8fc6f1a/src/Emanote/Model/Link/Rel.hs#L90-L97


> Not directly-related to this issue, but I see that relative markdown links like `[foo](../foo.md)` and [baz](./baz.md) are erroneously tagged as `RRTMissing`

Relative links are not supported yet. See https://github.com/srid/emanote/discussions/115#discussioncomment-1246389

> I really, really like the idea of decorating standard wiki links as `#[[Parent]]` and `[[Child]]#`; it seems to allow completely replacing tags with pages (i've called them "emboddied tags" in other formats like TiddlyWiki).

Yes, I've considered doing this in neuron before. There are some ideas here: https://github.com/srid/neuron/issues/379

I still think we can think of some new ways to "unify" tags and zettels in Emanote. Or at least come up with a way to associate (if not unify) hierarchical tags with zettels. But this should be a new Idea thread posted in Discussions.

> It would be cool if those parent/child links were indicated in the graph export,

Good idea. It is worth doing I think, for consistency if nothing else.

srid:
@edrex Can you give it one try now? Here are the changes:

- Back to `/-/export.json`
- I've flattened the structure so that the export now represents merely a list of files (you can build the graph yourselves however you want, including or not the parent folder, etc.; the export provides a 'low-level' data sufficient to do anything on the user side).
- Parent zettel is now part of the file meta (the "parentNote" key)
- Each file entry includes a "links" field that can be used to construct the graph
- Note that resolvedRelTarget refers to the generated URL (not source path, which the files list is keyed on), and you can build a reverse map of "url -> source path" using the files map.

srid:
Okay, I've merged this. Let's continue the discussion in [Discussions tab](https://github.com/srid/emanote/discussions) ... we still have opportunity to fine-tune the export before Emanote 1.0 release.

edrex:
I'm traveling for a few days, but I'll try out the new export functionality and follow up with any feedback when I have a chance.

PR #176, opened 2021-10-14
"edit this page" links

edrex:
Implements an "Edit this page" link, a common feature for file-based static site gens.

I thought it would be good for me to start learning the template/view
end of the codebase so I can help myself more.

It's incomplete, but I wanted to get feedback.

#### Issues

- [x] ~~Rather than choosing our own key (I used `site.editBaseUrl`) find                                 
what other tools use and follow/boost convention.~~ *Results: looking at hugo (not builtin) mkdocs.yml (concat repo_url with [edit_uri](https://www.mkdocs.org/user-guide/configuration/#edit_uri)), and mdbook (`edit-url-template` in [TOML](edit-url-template)), there doesn't seem to be a convention.*
- [ ] I just dropped the link in the footer without thinking about it
much. Where should it appear? Structurally, should it be in its own template?
- [x] Needs svg edit icon data. I borrowed the breadcrumb data. @srid what's
your process for generating these?
- [ ] I want to make this template contingent on a non-empty `baseUrl` meta key, but didn't see how to do that. Seems like Heist wants conditionals in Haskell, so I guess that requires some extra rendering logic?

### DIFF
--- a/default/templates/components/footer.tpl
+++ b/default/templates/components/footer.tpl
@@ -27,6 +27,16 @@
         </path>
       </svg>
     </a>
+  </div>
+  <div>
+    <a href="${ema:note:editUrl}" title="Edit this page" rel="edit">
+      <svg style="width: 1rem;" class="hover:text-${theme}-700" fill="none" stroke="currentColor"
+        viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
+        <path fill-rule="evenodd"
+          d="M7.293 14.707a1 1 0 010-1.414L10.586 10 7.293 6.707a1 1 0 011.414-1.414l4 4a1 1 0 010 1.414l-4 4a1 1 0 01-1.414 0z"
+          clip-rule="evenodd"></path>
+      </svg>
+    </a>
 
   </div>
 </footer>
--- a/default/templates/components/footer.tpl
+++ b/default/templates/components/footer.tpl
@@ -30,13 +30,11 @@
   </div>
   <div>
     <a href="${ema:note:editUrl}" title="Edit this page" rel="edit">
-      <svg style="width: 1rem;" class="hover:text-${theme}-700" fill="none" stroke="currentColor"
-        viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg">
-        <path fill-rule="evenodd"
-          d="M7.293 14.707a1 1 0 010-1.414L10.586 10 7.293 6.707a1 1 0 011.414-1.414l4 4a1 1 0 010 1.414l-4 4a1 1 0 01-1.414 0z"
-          clip-rule="evenodd"></path>
+      <svg width="16" height="16" viewBox="0 0 20 20" fill="currentColor">
+        <path
+          d="M13.586 3.586a2 2 0 112.828 2.828l-.793.793-2.828-2.828.793-.793zM11.379 5.793L3 14.172V17h2.828l8.38-8.379-2.83-2.828z">
+        </path>
       </svg>
     </a>
-
   </div>
 </footer>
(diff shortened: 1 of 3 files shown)


srid:
Quick response:

You can get svg icons from https://heroicons.com/

If the user doesn't specify editUrl in the yaml, I think we should just not show the edit link at all (no need for the default "edit/" hack, which wouldn't work anyway?).  Heist is not great at handling optional values (see https://github.com/snapframework/heist/issues/128 ) - but I wonder if we can workaround that by using lists and `<with>` (see how tags are rendered).

srid:
> I want to make this template contingent on a non-empty `baseUrl` meta key,

Under what situation would someone even want to do that? What I mean is: if you use `/edit` as editBaseUrl, and concatenate baseUrl with the editBaseUrl - you will have mysite.com/edit/... - but since Emanote generated website includes no folder "edit" it wouldn't work anyway.

edrex:
I meant "nonempty editBaseUrl". Yeah the /edit path isn't useful; i put it in as a placeholder until i figure out how to omit the edit link from the output in case editBaseUrl isn't set.

edrex:
I'm not sure when I'm going to have time to dig in to figure out the empty-editBaseUrl case, so feel free to finish this, otherwise I'll come back to it next week or so.

srid:
By the way, @edrex, do you know there is a way to achieve what you are doing here but without changing Emanote itself? Basically add your template to `templates/components/footer.tpl` of your notebook, and it will be used automatically instead of the Emanote default note. The template system + layers is not documented, but see here: https://github.com/srid/emanote/discussions/131

(You can just hardcode stuff in your .tpl; and no need to define any special .yml attribute)

edrex:
yeah, i just wanted to contribute something useful. I almost got this going yesterday, just needed to coax [lookupAeson](https://github.com/srid/emanote/blob/master/src/Emanote/Model/Note.hs#L223) to return a Maybe

srid:
@edrex I'm closing this in favour of #201 .. checkout the diff to see how you can use it on your notebook.

RoyiAvital:
This doesn't appear on the default `yaml`: https://github.com/srid/emanote/blob/master/default/index.yaml. Can you add an example?

Currently the EmaNote template makes the links go to the EmaNote repository.

**Update**: It seems to be part of `after-note.tpl`.  I find it much less friendly.

srid:
> **Update**: It seems to be part of `after-note.tpl`. I find it much less friendly.

You have just discovered the undocumented HTML templating system (a powerful feature that neuron doesn't support), which allows you to override/customize *any* of HTML/CSS that Emanote generates. It works this way: look for any folder/file under https://github.com/srid/emanote/tree/master/default and copy it to your notebook, then make modifications. 

`emanote-template` does it for https://github.com/srid/emanote/blob/master/default/templates/hooks/after-note.tpl - which is empty by default. It gets included here: https://github.com/srid/emanote/blob/master/default/templates/components/note-body.tpl#L4

srid:
These sites make use of HTML templating:

https://blog.hemera.systems/ (view [source](https://github.com/TheNeikos/hemera.systems))
https://ners.ch/ (view [source](https://github.com/ners/ners.ch))

RoyiAvital:
The template system is amazing.
But I think that the editing base URL should be a parameter in the YAML which is used by the template.

srid:
That's a good idea. Done in https://github.com/srid/emanote-template/commit/8d75688abc189fa43bcedb2d6c9a22677df52a07

edrex:
Nice, I need to get my public site up to exercise all this (still need to scrub private stuff, almost there). IIRC when I was looking at this there was some code forge whose edit URLs didn't have the form `|{editBaseUrl}/{note:sourcePath}`, but checking now at least Gitea (codeberg) and Gitlab do, so the complexity of making `editUrl` some sort of template string doesn't seem warranted.

Issue #153, opened 2021-09-24
Official release


srid:
Emanote is currently in beta. It is time to cut the first public release.

- [x] WIP -> Beta
- [ ] Finish [milestone tasks](https://github.com/srid/emanote/milestone/3)
- [x] Complete documentation
  - HTML templates
  - Layers
- [x] Release dependent libraries to Hackage
  - [x] heist (as a fork package, due to upstream being unmaintained)
  - [x] tagtree
  - [x] https://github.com/well-typed/ixset-typed/issues/17
  - [x] Ema https://github.com/srid/ema/pull/81
- [x] ~~(Optional) Build Linux static binary using Nix~~
  - [x] Try haskell.nix (including cross-compilation for musl & Windows) #155
- [x] Add to nixpkgs 
- [x] Reduce dependencies
  - [x] Move tests to its own stanza
  - [x] https://github.com/EmaApps/ema/issues/141
- [ ] Demo of Emanote as library
- [ ] Release emanote to Hackage
  - [x] First release
  - [x] Final release
- [ ] Update https://emanote.srid.ca/start/install to mention nixpkgs install (as default?)
- [ ] Blog post and annoucements

srid:
Not gonna bother with static builds, because we still have to figure out a way to distribute default layer along with it (cf. #156). Let package managers worry about it. We'll make emanote available on Hackage, so a simple 'cabal install emanote' would work (as long as non-Haskell dependencies, like tailwind, are installed).

srid:
Heist fork has been put in Hackage, https://github.com/snapframework/heist/pull/130#issuecomment-1086931247

srid:
Now that [Ema 0.8 is released](https://srid.ca/ema-0.8) it is high time we release Emanote to Hackage, just to get it to nixpkgs.

srid:
Emanote is feature-complete enough to make the initial release now.

srid:
Now on Hackage: https://hackage.haskell.org/package/emanote-0.8.0.0

I'd like to wait for a couple of things before making the blog post announcement:

- Add emanote to nixpkgs (perhaps @maralorn may be interested?)
- Add an example that shows how to use Emanote as a library to extend it (I have an idea for creating an invoice maker that uses Markdown tables and backlinks).

maralorn:
@srid What do you envision as "adding to nixpkgs"? We can just use `haskellPackages.emanote`, right?

But I will gladly enter myself as emanote maintainer and make sure that it builds.

[...]

PR #189, opened 2021-11-05
Add external link icon (#136)

### LINKED ISSUE #136: Add icon to non-wiki links
### Discussed in https://github.com/srid/emanote/discussions/135

<div type='discussions-op-text'>

<sup>Originally posted by **j-steinbach** September 21, 2021</sup>
This might be a customizable HTML/CSS thing, but Emanote should add icons to local and external links. 

Benefits: 
- I see exactly which link leads to an internal note/node that holds a meaningful conversation and which I can edit and have control of.
- I see which link leads me to an external page which moves me out of my ZK.

See [this](https://note.ema.srid.ca/resources/vscode) emanote page for example. It should be signaled that the first link (Visual Studio Code) is external, whereas the second link (emanote-template) is internal.

This is also how it is done on Wikipedia (just check the sources footer), with an icon that is similar to [this](https://fontawesome.com/v5.15/icons?d=gallery&p=2&q=external).</div>

kukimik:
This adds an icon to external links. I've checked it on Firefox 94.0 and Chromium 95.0.4638.69. Screenshots attached.

image

image

Mostly it looks ok, but look at the image (FAIR Pledge) at the bottom of the first screenshot: the icon is displayed in the next line.

So this is not  quite ready to get merged. Do you have any ideas how to handle the image case? I'm not even sure about the expected behavior: should we also have an icon if the link content is just an image? What if it has some text but ends with an image?

I'd also like to know your opinion whether adding the `emanote-external` class to external links is the right way to go.


srid:
About the FAIR pledge, yes that's a problem; not sure what the solution here is.

About the `emanote-external` class, that looks fine ... but we should allow the user to disable this styling if they wish. I've been meaning to split `pandoc.tpl` into separate files (another good-first-issue for Haskell if anyone wants to open a PR), so once that happens - uses can customize their link styles by creating a `pandoc/link.tpl` file.

Feedback on design: the external link icon should be of grey color (for de-emphasiz) and not of blue ($theme).

---

Is there anyway we can generalize this to also implement (clickable) header anchors? eg: https://markkarpov.com/tutorial/megaparsec.html#controlling-backtracking-with-try

image

It would need backend support, but assuming it exists ...

kukimik:
> Feedback on design: the external link icon should be of grey color (for de-emphasiz) and not of blue ($theme).

Ok. If you want it grey, then we do not need all this mask trickery and we can just use an image with a fixed color.

> Is there anyway we can generalize this to also implement (clickable) header anchors?

I don't think CSS can be used for this. Without backend support one could add some JS that adds an anchor after each header once the page is loaded, but I wouldn't call this a pretty solution.

However, if the following TODO in `pandoc.tpl` is done, then I think the anchor can just be inserted as part of the template?

```
<!-- TODO: Expand the above kind of overriding (full DOM control) to other AST nodes (below) -->
```

>  I've been meaning to split pandoc.tpl into separate files (another good-first-issue for Haskell if anyone wants to open a PR), so once that happens - uses can customize their link styles by creating a pandoc/link.tpl file.

I think that's a good idea. I know some Haskell, so I could help with this, but at the moment I don't have much free time. Where (in which modules) should this be implemented? I'm not familiar with the structure of emanote yet.

kukimik:
> Feedback on design: the external link icon should be of grey color (for de-emphasiz) and not of blue ($theme).

How about this? 

obraz

> About the FAIR pledge, yes that's a problem; not sure what the solution here is.

Fixed this with JS that disables the icon if the `<a>` element contains an image. This may need some tweaking to handle other situations (e.g. `<svg>` instead of ``). See the script at the bottom of `base.tpl`.

> About the emanote-external class, that looks fine ... but we should allow the user to disable this styling if they wish. I've been meaning to split pandoc.tpl into separate files (another good-first-issue for Haskell if anyone wants to open a PR), so once that happens - uses can customize their link styles by creating a pandoc/link.tpl file.

What do you think about also having a separate CSS file (or a set of files?) instead of putting the styles inside `base.tpl` (or inlining them in `pandoc.tpl`)?

srid:
Sorry I let this slip my radar.

srid:
> > Feedback on design: the external link icon should be of grey color (for de-emphasiz) and not of blue ($theme).
> 
> How about this?
> 
> obraz
> 
> > About the FAIR pledge, yes that's a problem; not sure what the solution here is.
> 
> Fixed this with JS that disables the icon if the `<a>` element contains an image. This may need some tweaking to handle other situations (e.g. `<svg>` instead of ``). See the script at the bottom of `base.tpl`.

I think I'd prefer the JS do the *opposite*. ie., start with no icon next to links, and have the JS add the icon to all text links (links with text, not image). It is safer approach, as we can't handle all edge cases.

> > About the emanote-external class, that looks fine ... but we should allow the user to disable this styling if they wish. I've been meaning to split pandoc.tpl into separate files (another good-first-issue for Haskell if anyone wants to open a PR), so once that happens - uses can customize their link styles by creating a pandoc/link.tpl file.
> 
> What do you think about also having a separate CSS file (or a set of files?) instead of putting the styles inside `base.tpl` (or inlining them in `pandoc.tpl`)?

Inline styles is fine for now. If they become huge, we can move it latter in a different PR.

---

@kukimik The screenshot you pasted looks good. Once the JS thing is addressed, we can merge it (after I test it locally).

kukimik:
> I think I'd prefer the JS do the opposite. ie., start with no icon next to links, and have the JS add the icon to all text links (links with text, not image). It is safer approach, as we can't handle all edge cases.

On the other hand, with the current approach you can easily turn off the external link icon manually in a particular case by adding a class to a single link. With JS activating the icon this is, of course, also possible, but I think it gets a little bit more complicated.

srid:
The generated site should work on browsers with JavaScript disabled. 

That's another -- and arguably the most important -- reason for the JS injecting icons after page load.

kukimik:
@srid  Please have a look at this. The links containing an image in the link text are now recognized before HTML is generated, so no JavaScript is needed. I'm not sure if `Heist.Extra.Splices.Pandoc.Ctx` is the best place to do it; it seemed to be the easiest one. If the user wants to manually disable the icon for some external link, they just need to add the `emanote-external-no-icon` class to the link. To completely disable the feature one has to override `base.tpl`.

Perhaps the code in `Ctx.hs` could be made shorter (this is somewhat FizzBuzz-like), and probably the feature should be documented, but I want to know your opinion before fixing the details.

Sorry for taking so long. Family reasons.

kukimik:
What do you think: should the icons appear in the backlinks section?

I also see I need to fix the colors to make them consistent with other icons around.

srid:
I'll take a look later this week. Been busy.

srid:
@kukimik Just took a look at this.

- For backlinks, that can be done as part of https://github.com/EmaApps/emanote/issues/187 (not necessarily in this PR) 
- Your `containsImage` logic doesn't work for wikilinks: `[![[disaster-girl.jpg]]](https://google.com)` (I see the external icon), though it does work for regular links `[](https://google.com)`. You can change the logic so that icon is disabled when link text contains *any* wiki-links (embed or not). 
- We should add tests for this logic (checks for plain text, images and wiki-links including embed ones); see `bin/test`. As well as, actually use these image links (both wikilink and normal) in the demo docs.

I can refactor the code latter.

srid:
Behaviorally, the icon looks good to me (and selecting text doesn't select it anymore). Maybe it should be made smaller. What do you mean by "fix the colors"?

kukimik:
> For backlinks, that can be done as part of https://github.com/EmaApps/emanote/issues/187 (not necessarily in this PR)

I believe it's sufficient to add the `emanote-external` classes in `context.tpl`. Looks like it works.

> Your containsImage logic doesn't work for wikilinks: [![[disaster-girl.jpg]]](https://google.com) (I see the external icon), though it does work for regular links [](https://google.com). You can change the logic so that icon is disabled when link text contains any wiki-links (embed or not).

I disabled the icon if the link contains another link. I believe this should detect wikilinks. Am I missing something?

> We should add tests for this logic (checks for plain text, images and wiki-links including embed ones); see bin/test. As well as, actually use these image links (both wikilink and normal) in the demo docs.

TBD

> and selecting text doesn't select it anymore

AFAIR this was an issue in #190, not here.

> Maybe it should be made smaller.

Made it a bit smaller.

> What do you mean by "fix the colors"?

I thought that the color for inactive icon should be lighter (like the color of the "Edit this page" icon), but I checked it and it doesn't look good - the icons are hardly visible. The current colors are ok.

srid:
Okay, behaviourally things look good to me 👍🏿 

Why do we need `NoIcon` differentiation at all? Can't we get rid of it?

kukimik:
> Why do we need NoIcon differentiation at all? Can't we get rid of it?

I'm not sure if I understand the question correctly.

External links without an icon still need to behave like other external links (e.g. `class="hover:underline" target="_blank" rel="noopener"` in `pandoc.tpl`). So we have external links with an icon and external links without an icon.

I see the following options to implement this:
* For external links inject or omit  the `emanote-external` class directly in the code, without using templates. (Or maybe we can have two classes: `emanote-external-icon` and `emanote-external-no-icon`, and inject the right one. Or use some other variant of this approach. The main point is that we do it in `.hs`, not in `.tpl`)
  * This is against the tendency to keep styling in the templates. But as long as the classes are described in the docs, I'm ok with it.
* Keep it all in the templates. That's what I did.
  * I believe that this way the `emanote-external*` classes may seem less "magical" to the end user, and may be understood  by just looking at the templates, without reading the docs and without reading Haskell.
  * It also allows the user to easily apply some other tailwind styling when the icon can't  be used. But probably this extra flexibility is not needed.
  * On the other hand, it clutters the templates.

There are some mixed solutions, but I don't find them very clean.

 If you find the first approach better, then I may rewrite it. It should be easy.

srid:
@kukimik Basically, I meant removing the `NoIcon` thing entirely. I just tried it out and it works. See https://github.com/EmaApps/emanote/commit/9dee6147e5529beb1f6461915d69909781e9c5dd

Do you see any issue with this change? I cannot think why users would want to customize the "no icon" case. Can you point out a real-world example?

kukimik:
> Do you see any issue with this change?

The external links containing an image (or a wikilink) lose the `target='_blank'` and `rel='noopener'` attributes. These attributes are set by:

```
<External class="hover:underline emanote-external" target="_blank" rel="noopener" />
```

That's what I meant here:

> External links without an icon still need to behave like other external links (e.g. class="hover:underline" target="_blank" rel="noopener" in pandoc.tpl). So we have external links with an icon and external links without an icon.

On the other hand this made me think some more about my proposed solution and now I think that the following structure would be better:

```
<External [no extra class here]>
  <WithIcon class="emanote-external-icon">
</External>
```
This, together with an extra `emanote-external-no-icon` class and a minor change in `Ctx.hs`, would allow the user to easily force showing/hiding the icon for particular links. (Previously they were only easy to hide, but I think forcibly showing them is a more common case, e.g. I'd like to be able to force an external icon for `<a href='...'>A link text containing some  in the form of an inline image</a>`.

------------

>  I cannot think why users would want to customize the "no icon" case. Can you point out a real-world example?

I don't have such an example. That's why I said "probably this extra flexibility is not needed."

srid:
Let's take a step back and flesh out the requirements first.

Question 1: What are the specific Markdown snippets on which we do *not* want to add the external icon? So far, one has been identified:

1. An image wrapped in a link (see the disaster girl image [here](https://emanote.srid.ca/demo/embed)). (Note: this doesn't include your "A link text containing some image" case).

Is there anything else? I'm not sure if we should bother with wiki-links since that is an absurd case (is there a legitimate reason for them?).

Question 2: Why do we need more than one template class var, `External`, here? This is my original question: why the separate WithIcon/WithNoIcon thing? Let's see some specific examples in Markdown. It may even be better to use them in ./docs/demo.

srid:
> An image wrapped in a link

Actually, this is only valid if the image appears as a block element (as with the disaster girl, due to being the only element of a `<p>`). Which makes the requirement even more annoying to property define.

EDIT: But maybe the inline icon is undesirable even in inline images?

srid:
I'm mostly envisioning a situation where the user doesn't *have to* override their .tpl - tweaking WithIcon/NoIcon just to workaround the external link icon behavior. It should do the "right" thing as default.

srid:
Yea, this is really frustrating to think about without having *specific* and *concrete* examples. I'd suggest adding all those examples (including your "A link text containing some image ..." case) to `./docs/demo` in some form first. It could just be a sub-heading on the 'Extended Markdown' page.

Then defining the requirements becomes straightforward, and it is easy to talk about this from a common ground.

kukimik:
This is going to be a heuristic. A better or worse one, but still. I don't believe we can cover all the usecases. I'm trying to illustrate this in the `custom_styling.md` section. I propose to show the icon by default if the link contains some text, and hide it otherwise.
This can be easily overridden by supplying an appropriate class.

I also propose an implementation that I find satisfying. I think it's pleasantly symmetric and readable at the cost of slight redundance.

No tests yet. I've not looked into this.

kukimik:
I've noticed that I duplicate the `emanote-external-icon`/`emanote-external-no-icon` classes if they are set by hand. Are you ok with changing `<>` to `Data.List.union` in this line? https://github.com/EmaApps/emanote/blob/d2ae1013094537f76692edc910bf36b2fb543107/src/Heist/Extra/Splices/Pandoc/Attr.hs#L21

docs/demo/custom-style.md
```diff
@@ -46,3 +46,19 @@ Using [fenced_divs](https://github.com/jgm/commonmark-hs/blob/master/commonmark-
...
+
+* No special styling:
+  * [the unicode.org website](https://www.unicode.org/)
+  * [why the external link symbol ![[external-link-icon.svg]] is not in Unicode](https://www.unicode.org/alloc/nonapprovals.html)
+  * [![[hello-badge.svg]]](https://emanote.srid.ca)
+* `{.emanote-external-no-icon}` used to suppress the icon:
+  * [the reasons behind Unicode rejecting the external link symbol ![[external-link-icon.svg]]](https://www.unicode.org/alloc/nonapprovals.html){.emanote-external-no-icon}
+  * A 90's style hyperlink:
+    :::{.flex .justify-center}
+    [➡➡➡ **CLICK HERE!!!** ⬅⬅⬅](https://emanote.srid.ca){class="shadow-lg border-8 rounded-md bg-yellow-400 border-red-200 emanote-external-no-icon"}
```
srid: wouldn't this make more sense?

```diff
--- a/docs/demo/custom-style.md
+++ b/docs/demo/custom-style.md
@@ -56,7 +56,7 @@ Emanote displays an icon next to external hyperlinks, provided that the link con
   * [why the external link symbol ![[external-link-icon.svg]] is not in Unicode](https://www.unicode.org/alloc/nonapprovals.html)
   * [![[hello-badge.svg]]](https://emanote.srid.ca)
 * `{.emanote-external-no-icon}` used to suppress the icon:
-  * [the reasons behind Unicode rejecting the external link symbol ![[external-link-icon.svg]]](https://www.unicode.org/alloc/nonapprovals.html){.emanote-external-no-icon}
+  * [the reasons behind Unicode rejecting the external link symbol](https://www.unicode.org/alloc/nonapprovals.html){.emanote-external-no-icon}
   * A 90's style hyperlink:
     :::{.flex .justify-center}
     [➡➡➡ **CLICK HERE!!!** ⬅⬅⬅](https://emanote.srid.ca){class="shadow-lg border-8 rounded-md bg-yellow-400 border-red-200 emanote-external-no-icon"}
```

srid:
@kukimik I'm happy with this PR. If you could add tests for `determineLinkType` (which contains the heuristic), I can look into merging it after a cleanup. See the various `*Spec.hs` files for comparison.

srid:
> Are you ok with changing `<>` to `Data.List.union` in this line?

Yea, that makes sense.

kukimik:
Thanks for pushing me to write the tests. While writing them I noticed that the `Code` and `Math` constructors should also be treated as text.

I'm not sure what is the `RawInline`. Ignored it for now. I think that `Cite` is not used in emanote, and I also ignored it.

The `Footnote` constructor is also not in the tests. It may break the heuristic (because the string may appear in the body of the footnote), but on the other hand, inserting a footnote in the link text results in broken HTML (nested `<a>` tags) and makes little sense.

kukimik:
I'll try to address #348 somehow. Perhaps it won't be the ultimate solution, but I think I've got an idea.

kukimik:
@srid Can you have a look at the new version? I'll probably have to correct the wording in `custom_styles.md`, include a few more tests and maybe improve the look of the icons. But, forgetting about the details, what do you think about this approach?

src/Heist/Extra/Splices/Pandoc/CtxSpec.hs
```diff
@@ -0,0 +1,38 @@
...
+import Relude
+import Test.Hspec
+import Test.Hspec.Hedgehog
+import Text.Pandoc.Definition
+
+spec :: Spec
+spec = do
+  describe "determineLinkType" $ do
+    it "recognizes internal URLs" . hedgehog $ do
+      determineLinkType [Str "example", Space, Str "link"] [] "/some/example/path#id" === LinkType_Internal
```
srid: I'd like to see this test made more readable. Instead of passing a parsed AST, just pass the raw markdown, viz.:

```haskell
determineLinkTypeFromMarkdown "[example link](/some/example/path#id)" === LinkType_Internal
```

docs/demo/custom-style.md
```diff
@@ -46,3 +59,25 @@ Using [fenced_divs](https://github.com/jgm/commonmark-hs/blob/master/commonmark-
...
+* No special styling:
+  * [the emanote repo](https://github.com/EmaApps/emanote)
+  * [why the external link symbol ![[external-link-icon.svg]] is not in Unicode](https://www.unicode.org/alloc/nonapprovals.html)
+  * [$e^{i \pi} + 1 = 0$](https://en.wikipedia.org/wiki/Euler%27s_identity)
+  * [`(>>=) :: forall a b. m a -> (a -> m b) -> m b`](https://hackage.haskell.org/package/base/docs/Prelude.html#v:-62--62--61-)
+  * [![[hello-badge.svg]]](https://emanote.srid.ca)
+  * mother_hetchel@weaversguild.com
+  * [+12 3456789](tel:+123456789)
+* `{.emanote-external-no-icon}` used to suppress the icon:
+  * Water's formula is [H](https://en.wikipedia.org/wiki/Hydrogen){class="emanote-external-no-icon"}₂[O](https://en.wikipedia.org/wiki/Oxygen){class="emanote-external-no-icon"}.
```
srid: `{class="emanote-external-no-icon"}` seems too verbose a way to disable the external icon. How about `{data-linktype=none}` or something like that instead? So the CSS query would be something like `a.emanote-external-icon:not([data-linktype/=none])` (assuming a syntax like that is valid).

src/Heist/Extra/Splices/Pandoc/Ctx.hs
```diff
@@ -115,16 +121,75 @@ inlineLookupAttr node = \case
...
+        pure $
+          attrFromNode linkNode
+            `concatAttr` attrFromMaybeNode linkTypeNode
+            `concatAttr` attrFromMaybeNode iconNode
+
+data LinkType = LinkType_Internal | LinkType_External ExternalLinkIconStatus
+  deriving stock (Eq, Show)
+
+data ExternalLinkIconStatus = Icon | NoIcon
+  deriving stock (Eq, Show)
```
srid: I'm curious why you went with two different types. Why not instead this:

```haskell
-- Regular links do not have any special styling (like icons)
data LinkType = LinkType_Regular | LinkType_Internal | LinkType_External
```

This way, I imagine users can explicitly choose them in CSS style using `[foo](mailto:s@s.com){data-linktype=regular)`.

srid (commented):
Semantically, this looks good. But I think we want to improve the user-facing "API" (the class/attrs) a bit.

src/Heist/Extra/Splices/Pandoc/CtxSpec.hs
kukimik: Ok. I didn't want to involve the parser in order to keep the test "local" at the cost of readability. I'll improve that when I find another bit of free time. (Currently I can manage to spend just like 1hr a week on coding for fun...)

docs/demo/custom-style.md
kukimik: > `{class="emanote-external-no-icon"}` seems too verbose a way to disable the external icon. 

`{.emanote-external-no-icon}` is a bit shorter. The class name could also be shortened, of course. And in principle this is a workaround that shouldn't be used much, as long as the heuristic is good.

> How about `{data-linktype=none}` or something like that instead?

Nice idea. I hadn't thought about using the data-attributes. These are no longer booleans, which has its good and bad sides. I used the data attributes in the new version.

src/Heist/Extra/Splices/Pandoc/Ctx.hs
kukimik: > I'm curious why you went with two different types. Why not instead this:

1. Because I know what an internal link is (a link pointing to an internal resource of the site I'm visiting) and what an external link is (a link pointing to an off-site resource). And I know what is an external link with an icon and an external link without an icon. But I have no clue about a regular (or a [normal](https://github.com/EmaApps/emanote/discussions/348#discussioncomment-3609432)) link. (And the name suggests these should be the prevalent ones). For me the icon thing is about the styling, not about the link type.
2. I believe I do have three types after all: Internal, External with an icon, External without an icon.
3. Because in `pandoc.tpl` I would have to repeat the `target="_blank" rel="noopener"` attributes for the *regular* links, and I didn't like the repetition:

```
  <PandocLink class="text-${theme}-600">
    <Internal class="mavenLinkBold hover:underline" />
    <External class="hover:underline" target="_blank" rel="noopener" />
  </PandocLink>
```

> This way, I imagine users can explicitly choose them in CSS style using [foo](mailto:s@s.com){data-linktype=regular).

I used the `data` idea.

kukimik:
Here's a new version.

When `Heist.Extra` moved to a separate library, I began to doubt whether `Heist.Extra.Splices.Pandoc.Ctx` is th best place to do the job. I moved it to a builtin filter. I'd like to know your opinion.

On the other hand, in `Heist.Extra.Splices.Pandoc.Ctx` links with an URI scheme should be recognized as external (currently only links containing `://` in the address are recognized as external) to keep things consistent, so the logic in

```
    hasURIScheme =
      isRight . P.parse schemeP ""
      where
        schemeP = do
          c <- PC.letter
          cs <- P.many (PC.alphaNum P.<|> P.oneOf ".-+")
          _ <- PC.char ':'
          return (c : cs)
```

should be included there somehow. Duplication? Common dependency? Adding something like [network-uri](https://hackage.haskell.org/package/network-uri-2.7.0.0) to the dependencies list seems an overkill for just this snippet.

-------------

I also used your idea on data-attributes, which simplified the code, removed the custom icon for the `tel:` scheme (I believe the custom styling for `mailto:` is sufficient both in most usecases and as an example) and improved the icons. 

No tests yet (see one of the comments).

kukimik:
I just force-pushed a rebase, the tests and minor stylistic changes in `custom-style.md`.

kukimik:
Huh.. I've just found out that the builtin filters are not applied to the note excerpts in the backlinks ("Links to this page") section. This means that with the current approach the icons won't be displayed in this section. :disappointed: On the other hand, this also means that tags are not converted to links and that the emoji don't get their `font-family: emoji` styling. Is this behavior intended? It looks like a bug.

kukimik:
@srid I have an idea for an improvement and a working prototype; should deliver it in the next few days. Please don't read the PR in its current state. Sorry, I don't feel very comfortable with this. :confused:

kukimik:
> I have an idea for an improvement

Finally I decided against it; I just made some minor changes. I don't think I can make it much better anytime soon. Please have a look when you have some time. Looks like it works, maybe it'll be enough.

The `data-linkicon` "API" is designed to be extendable, so that we may for example support `data-linkicon=filetype` that displays a small PDF icon next to links to a PDF file, or `data-linkicon=anchor` for #137, and so on.

Most of the styling could probably be done in Tailwind instead of pure CSS, but I think the `after:` variant modifiers are not supported (at least in the version of Tailwind used in the live preview).

docs/guide/markdown/custom-style.md
```diff
@@ -1,10 +1,23 @@
...
+    <script>
+      window.MathJax = {
+        startup: {
+          ready: () => {
+            MathJax.startup.defaultReady();
+          }
+        }
+      };
+    </script>
+    <script async="" id="MathJax-script" src="https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-mml-chtml.js"></script>
```
srid: Replace all of this with `<snippet var="js.mathjax" />` (see `math.md`).

docs/guide/markdown/custom-style.md
```diff
@@ -46,3 +59,28 @@ Using [fenced_divs](https://github.com/jgm/commonmark-hs/blob/master/commonmark-
 
 
 [^mob]: If you are viewing this page on mobile or smaller screens, the embedded notes will be stacked on top of one another because we use Tailwind's [responsive classes](https://tailwindcss.com/docs/responsive-design). Incidentally, we use the `{class=".."}` syntax, rather than the `{.someClass}` syntax, only because the former is [more lenient](https://github.com/jgm/commonmark-hs/issues/76) in accepting non-standard class names, such as the Tailwind responsive classes (eg. `lg:grid-cols-2`).
+
+## External link icons
```
srid: Wherever you refer to `template` in this section, you probably want to rename it to `[[html-template|HTML template]]`.

src/Emanote/Pandoc/BuiltinFilters.hs
```diff
@@ -51,3 +54,35 @@ fixEmojiFontFamily =
...
+      | hasURIScheme url && containsText inlines ->
+        let showLinkIconAttr = ("data-linkicon", "external")
+            newAttrs = insert attrs showLinkIconAttr
+         in B.Link (id', classes, newAttrs) inlines (url, title)
+    x -> x
+  where
+    insert as a
+      | fst a `elem` (fst <$> as) = as
+      | otherwise = a : as
+    hasURIScheme =
```
srid, line 80: Instead of writing your own URI parser, I'd just use the [modern-uri](https://hackage.haskell.org/package/modern-uri-0.3.6.0/docs/Text-URI.html) package. Its [`URI`](https://hackage.haskell.org/package/modern-uri-0.3.6.0/docs/Text-URI.html#t:URI) type already breaks down the URI components, including the scheme.

src/Emanote/Pandoc/BuiltinFilters.hs
```diff
@@ -51,3 +54,35 @@ fixEmojiFontFamily =
...
+      isRight . P.parse schemeP ""
+      where
+        schemeP = do
+          c <- PC.letter
+          cs <- P.many (PC.alphaNum P.<|> P.oneOf ".-+")
+          _ <- PC.char ':'
+          return (c : cs)
+    containsText =
```
srid, line 91: Please add a type-signature and, more importantly, comments as to what this function does.

src/Emanote/Pandoc/BuiltinFiltersSpec.hs
```diff
@@ -0,0 +1,48 @@
...
+import Test.Hspec.Hedgehog (hedgehog)
+import Text.Pandoc.Definition (Inline (..))
+import Text.Pandoc.Walk (query)
+
+spec :: Spec
+spec =
+  do
+    describe "setExternalLinkicon" $ do
+      let run =
+            (>>=) . evalEither . fmap (query extractAttr . preparePandoc . snd) . parseMarkdown ""
```
srid: I have a hard time understanding this code with a casual glance. Could you rewrite it verbosely without using function composition and the like?

src/Emanote/Pandoc/BuiltinFiltersSpec.hs
```diff
@@ -0,0 +1,48 @@
...
+import Test.Hspec
+import Test.Hspec.Hedgehog (hedgehog)
+import Text.Pandoc.Definition (Inline (..))
+import Text.Pandoc.Walk (query)
+
+spec :: Spec
+spec =
+  do
+    describe "setExternalLinkicon" $ do
+      let run =
```
srid: Why is this called "run"? What does this function do? What does it return? You want to clarify all that in some fashion.

src/Emanote/Pandoc/BuiltinFiltersSpec.hs
```diff
@@ -0,0 +1,48 @@
...
+            (>>=) . evalEither . fmap (query extractAttr . preparePandoc . snd) . parseMarkdown ""
+            where
+              extractAttr (Link (_, _, attrs) _ (_, _)) = snd <$> filter ((== "data-linkicon") . fst) attrs
```
srid: Should this be called `getDataLinkIconAttr` instead?

srid (changes requested):
Did an initial review. The functionality looks great. The tests need improvement. I haven't reviewed the test body itself, because I do not understand the semantics of it (eg.: what are all those `=== []`??); see my comment on `run`.

srid:
By the way, this feature _could_ go in #153 if we can get it ready before the release blog post. I can always make further minor releases to Hackage.

kukimik:
> By the way, this feature _could_ go in #153 if we can get it ready before the release blog post. I can always make further minor releases to Hackage.

Thanks for the review. I'll try to do it over the weekend, but can't promise anything.

kukimik:
> Thanks for the review. I'll try to do it over the weekend, but can't promise anything.

Sorry, I didn't manage to do it.  Don't wait for me. I think this may be released later.

src/Emanote/Pandoc/BuiltinFiltersSpec.hs
kukimik: Fixed.

src/Emanote/Pandoc/BuiltinFiltersSpec.hs
kukimik: I hope it looks better now.

src/Emanote/Pandoc/BuiltinFiltersSpec.hs
kukimik: I hope it looks better now.

src/Emanote/Pandoc/BuiltinFilters.hs
kukimik: Fixed.

docs/guide/markdown/custom-style.md
kukimik: Fixed.

src/Emanote/Pandoc/BuiltinFilters.hs
kukimik: I'd rather not use `modern-uri` for this.
1. It adds an extra dependency for something very simple.
2. If I understand correctly, it only parses URIs that are valid as per RFC 3986. I'm not sure, but I think this document [may already be obsolete](https://github.com/mrkkrp/modern-uri/issues/39#issuecomment-805877306), and who knows what happens in the future? The web is full of non-standard-conforming code, and I don't really need the whole URI to be correct in order to extract just the scheme component. Unfortunately `modern-uri` does not export the [`pscheme`](https://hackage.haskell.org/package/modern-uri-0.3.6.0/docs/src/Text.URI.Parser.Text.html#pScheme) parser. I don't know if any other of the URI-related libraries exports the scheme parser. I think I haven't seen one. Maybe the maintainer of `modern-uri` (or some other lib) may agree to expose the scheme parser?

On the other hand, I'd like to have the parser available both in https://github.com/srid/heist-extra (see https://github.com/srid/heist-extra/blob/7ca33a42564990df6fa92267626a17d788bf65a1/src/Heist/Extra/Splices/Pandoc/Ctx.hs#L120) and probably also https://github.com/srid/commonmark-wikilink (see https://github.com/srid/commonmark-wikilink/blob/72c880b5dbe8521f8c424c7bd39cb11b26e8dbb8/src/Commonmark/Extensions/WikiLink.hs#L106) to make things uniform. So it would be nice to have some external source for it.

docs/guide/markdown/custom-style.md
kukimik: Fixed. But in fact I followed the instructions in `math.md` when adding this. It's a little confusing. I'll correct `math.md` in another PR to make it uniform with `mermaid` and `syntax-highlighting`. (Maybe there should be a separate page about JS snippets?)

kukimik:
Thanks again for the review.

> The functionality looks great

Good to hear that. I got somewhat frustrated with this. It looked easy at the beginning, but there is no obvious "right" solution here.

kukimik:
Rebased to master to run treefmt using new Ormolu.

docs/guide/markdown/custom-style.md
srid: Yea, we should improve the docs to state both the 'raw' and the 'alias' version. Separately, not in this PR.

default/templates/base.tpl
```diff
@@ -23,13 +23,41 @@
...
+    a[data-linkicon]:not([data-linkicon=""]):not([data-linkicon="none"])::after {
+      /* filter converts black to rgb(156,163,175) */
+      filter: invert(71%) sepia(3%) saturate(904%) hue-rotate(179deg) brightness(92%) contrast(87%);
+      margin-left: 1px;
+    }
+
+    a[data-linkicon]:not([data-linkicon=""]):not([data-linkicon="none"]):hover::after {
+      /* filter converts black to  rgb(175,85,99) */
+      filter: invert(32%) sepia(10%) saturate(834%) hue-rotate(176deg) brightness(92%) contrast(88%);
+    }
```
srid, line 44: Do we know what the two clauses above do? Having a comment explaining the purpose of these styles would be useful.

srid (approved):
I think this is ready to be merged! ~~Just one thing I don't understand in base.tpl.~~

src/Emanote/Pandoc/BuiltinFiltersSpec.hs
srid: It is :-) I've also refactored it a bit more to satisfy my subjective view on clarity.

default/templates/base.tpl
srid: Oh n/m. This just overrides the default black covert for default/hover states.

kukimik:
@srid Thanks for the final corrections. And for your guidance throughout the process. And for your patience - the PR was opened over a year ago! Glad to see it merged.

PR #202, opened 2021-11-28
Switch to a variable font

### LINKED ISSUE #197: Fonts don't look as good on macOS
image

srid:
Resolves #197


srid:
Could have tweaked more, cf. https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Fonts/Variable_Fonts_Guide

PR #214, opened 2021-12-09
fix folgezettel cycle detection

### LINKED ISSUE #210: modelFolgezettelAncestorTree doesn't terminate with reciprocal parent links
I'm seeing runaway CPU usage running the server or gen with a test case consisting of an `index.md` linking with `#[[foo]]` to `foo.md` (https://gist.github.com/2e5117e799c17d5363f302615489195f)

 Doesn't happen if
  - if `foo.md` doesn't exist
  - with `index.md` renamed to `i.md`
  - with a non-parent link `[[foo]]`

Reproducible? I see it with builds of emanote-0.3.10.3 and emanote-0.3.15.3 from nix (where do those versions come from?)

edrex:
insert into visited set before recursion
fixes #210

### DIFF
--- a/emanote.cabal
+++ b/emanote.cabal
@@ -1,6 +1,6 @@
 cabal-version:      2.4
 name:               emanote
-version:            0.4.0.0
+version:            0.4.1.0
 license:            AGPL-3.0-only
 copyright:          2021 Sridhar Ratnakumar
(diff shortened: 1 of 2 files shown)


srid:
Great! Could you up the version in `emanote.cabal` (to `0.4.1.0`)? Then I can merge.

PR #215, opened 2021-12-11
improve layer option help text

edrex:
- Mention path delimiter
- correct description of overlay precedence (left is higher)
- example


srid:
> * correct description of overlay precedence (left is higher)

wait, what. if that's true, then the order of layers here is wrong:

https://github.com/srid/emanote-template/blob/master/.github/workflows/publish.yaml#L15

edrex:
Hmm, index.yaml seems to prioritize the later layers, but content seems to prioritize earlier layers (test: `.deploy/github/index.md is ignored`). So there's a bug.

Issue #188, opened 2021-11-04
Extend Emanote in Haskell


srid:
Allow Emanote to be used as a Haskell library (ideally via Nix) so that users can "extend" it with custom features. 

For example, I would like to add a "special" page to my notebook that is computed on the fly (to show tasks based on a non-general private context, display high-level goals/perspective, current-date/time-specific notes, ... a 'dashboard' of some sort), depending on the notebook contents. One could use `export.json` to customize things, but having raw direct access to the Emanote engine would allow us to do more. 

It should be possible to add new routes, and specify how to render them, down to the HTML (nothing is hardcoded).

Bonus points if we can do this all at runtime (ie. interpret Haskell code), but that's a stretch goal for a latter project.

srid:
> It should be possible to add new routes, and specify how to render them, down to the HTML (nothing is hardcoded).

Alternative approach: an user-created process that uses Emanote as a library to observe model changes, and generates a `dashboard.md` or whatever with generated Markdown (or HTML, for complex layouts), which file is picked up by Emanote live server (main process). The advantage of this approach here is we don't have to extend the routing mechanism (ie., no new core use case needed), and that the user 'adds' pages by simply adding a new note to the zettelkaten (ie. an existing core use case).

What's common to both approaches is the need to make Emanote available as a Haskell library, with the corresponding entry points exposed.

srid:
Heck, both approaches can be *combined* into a singular one - wherein this user-created process generates its special notes on a *separate layer*, and then runs a live-server with the main layer and the generated layer *merged* together. Separation of concerns is the advantage here. This is one of the scenarios the "layers" feature (core-use-case^1) enables.

^1 See `#architecture-and-design` in [fpslack](https://fpchat-invite.herokuapp.com/)

srid:
Closing in favour of the much better approach (due to decoupling) #247

PR #219, opened 2021-12-27
Emanote as a library

srid:
for writing of Emanote-like apps (cf. `Emanote.emanate`)

and as a foray into #188


srid:
Example: https://github.com/srid/banyan/commit/869991888349190855b3c963493f9ff352d250d0

PR #242, opened 2022-01-30
WIP: Live server - search

### LINKED ISSUE #241: Interactive search in live server
In live server, add a search box where users can search notes interactively.

This can be implemented one top of https://github.com/srid/ema/issues/69 

- [x] `ema.switchRoute`
- [ ] ema: support for query in URLs (ie., `/search?q=foo+bar`)
  - That, or we can get by without queries, ie., `/search/foo+bar`

srid:
Resolves #241 

- [x] Add route and render the query listing
- [ ] Add input box and JS code to call `ema.switchRoute` automatically (?)
- [ ] Cleanup

### DIFF
--- a/default/templates/special/queryview.tpl
+++ b/default/templates/special/queryview.tpl
@@ -1,0 +1,31 @@
+<apply template="/templates/special/base">
+  <bind tag="special-main">
+    <div class="w-full bg-gray-300">
+      <div class="w-full px-3 py-1">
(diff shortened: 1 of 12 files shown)


srid:
I won't be getting to this any sooner. If anybody else is interested, feel free to take it up!

srid:
Perhaps we should ditch this entirely and consider supporting client-side search with a server-generated index: https://stork-search.net/ - as it would also work equivalently on the static site.

PR #273, opened 2022-03-31
add cache substituter to flake

edrex:
Continued from https://github.com/srid/emanote-template/pull/4

Tested with direnv on NixOS system with `nix.settings.trusted-users = [ '@wheel' ];`

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,5 +1,9 @@
 {
   description = "emanote";
+  nixConfig = {
+    extra-substituters = "https://srid.cachix.org";
+    extra-trusted-public-keys = "srid.cachix.org-1:MTQ6ksbfz3LBMmjyPh0PLmos+1x+CdtJxA/J2W+PQxI=";
+  };
   inputs = {
     ema.url = "github:srid/ema/multisite";
     nixpkgs.follows = "ema/nixpkgs";


srid:
What happens if the current user is not trusted?

edrex:
> What happens if the current user is not trusted?

There's a warning message and the cache is skipped. 

[CA store paths](https://discourse.nixos.org/t/content-addressed-nix-call-for-testers/12881) will enable non-trusted users to add paths to the store, including substituters (as @balsoft explained to me)

srid:
Could you paste the console output showing those warnings?

srid:
Also, as I understand this works only for `nix develop` and `nix build`, but not `nix run`: https://github.com/NixOS/nix/issues/6170

What about `nix-env -if` or when using Emante as flake input? Does the cache gets used?

edrex:
> Could you paste the console output showing those warnings?

I'd have to set up an untrusted user to reproduce the warning, but here's the source line: [nix/src/libstore/daemon.cc#L225](https://github.com/NixOS/nix/blob/8ad485ea893862029e02cb560a15fd276753b04f/src/libstore/daemon.cc#L225).

Hopefully support for `nix run` will be added soon.

> What about nix-env -if or when using Emante as flake input? Does the cache gets used?
For `nix-env -if`, I haven't tested as I live in a post-flakes world. As an input, the settings don't apply, see https://github.com/NixOS/nix/pull/4189#issuecomment-717067395

srid:
Looks like this is currently only useful for caching development shell. Better than nothing, I guess.

srid:
I wonder how https://github.com/srid/emanote/commit/96c00336fc7bb8648c2d3cf371d4e1e3fe2c22d9 impacts this.

Issue #263, opened 2022-03-10
Pandoc lua filters support


srid:
Emanote definitely should support [lua filters](https://pandoc.org/lua-filters.html) so that users can include these filters in their notebook along with .md and .tpl files.

You should be able to apply these filters on per-note or per-folder basis via specifying them in .yml files. Just like you can specify a .tpl template.

_Originally posted by @srid in https://github.com/srid/emanote/discussions/250#discussioncomment-2332791_

---

- [x] MVP version: https://github.com/srid/emanote/pull/278
- [ ] Final version

srid:
We can call `Text.Pandoc.Filter.Lua.apply` in the patch function, but this is not enough, because updating the filters themselves won't trigger the re-patching of all Pandoc documents. 

Alternatively, we can apply the filter during the rendering stage - but `apply` is an IO action, and this will be expensive. It still won't solve the problem of automatically updating dependent docs (like backlinks).

Requires some thought here ...

srid:
Proof of concept: https://gist.github.com/srid/6e17e3ad68a3f6ba877a8024b8ea0909 w/ `git clone https://github.com/pandoc/lua-filters.git ./default/lua-filters`.

That gave me an idea that we should include https://github.com/pandoc/lua-filters by default in the 'default' layer. EDIT: But this can only be done after adding a mechanism to exclude folders from being copied over to the static site.

srid:
https://x.com/sridca/status/2052098639565094949

PR #278, opened 2022-04-18
[MVP] Basic Pandoc lua filters

srid:
A start at #263 

**Goal**: Add very basic support for [Pandoc lua filters](https://pandoc.org/lua-filters.html) in this PR, so that we can start playing with it from today.  Advanced functionality will be implemented in a later PR at a later time (after user feedback, and further architectural thinking).

Limitations:
- Lua file path can be specified in frontmatter only. 
  - In the advanced version, we should support yaml data cascading.
- Lua file path ~~is resolved relative to the current directory~~ should be ~~relative to the highest precedent layer (notebook path, typically).~~
  - [x] In the advanced version, we should look it up in all layers (just as we do for Heist template files). 
- Changes to the Lua file don't necessarily hot-reload the model. Workaround: `touch` the .md file in question. 
  - In the advanced version, this should just work without any workarounds.

Resolving all of the above is outside of the scope of this PR. Doing them will require us to think about dependency handling in the model, which is complex to implement and needs to be thought about carefully. This will all be addressed in the advanced version.

This PR should allow people to transform their Markdown notes dynamically in Lua, experiment willy-nilly and give early feedback to shape final Lua support in Emanote.

Test with filters here: https://github.com/pandoc/lua-filters

### DIFF
--- a/default/index.yaml
+++ b/default/index.yaml
@@ -50,7 +50,7 @@ pandoc:
     # in the generic class.
     emanote:inline-tag:a/red/tag: bg-red-100
     emanote:placeholder-message: text-gray-400 border-t-2 inline-block pt-0.5
-    emanote:error: text-l bg-red-200 p-2 border-2 border-black m-2 font-mono
+    emanote:error: text-l bg-red-100 p-2 border-2 border-black m-2 font-mono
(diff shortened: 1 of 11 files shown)


srid:
To run this PR directly without installing it,

```
nix run github:srid/emanote/lua-filters
```

PR #286, opened 2022-05-23
Prepend doctype to HTML document

### LINKED ISSUE #216: No DOCTYPE in generated HTML
**Describe the bug**
The generated HTML filles lack the `<!DOCTYPE html>` declaration.

**To Reproduce**
1. Follow the instructions at https://note.ema.srid.ca/start/install/docker.
2. Open the generated HTML files in your favourite text editor and check the first line.

The pages published at https://note.ema.srid.ca/ are also affected by this bug.

**Expected behavior**
The generated HTML files should start with the following text:

`<!DOCTYPE html>`

- Emanote 0.3.15.2 used via docker on Linux.

**Additional context**

About the DOCTYPE: https://developer.mozilla.org/en-US/docs/Glossary/Doctype

Note that the text `<!DOCTYPE HTML>` is present in `base.tpl`. I don't know why it is not included in the generated HTML.

I believe this should be fixed before the 1.0 release.

ners:
This is a quick fix until the matter is investigated more thoroughly. Allowing the browser to go into quirks mode causes all kinds of breakages, e.g. KaTeX.

My cursory analysis suggests that the doctype is discarded when a template is parsed, as any other HTML comment in Pandoc's intermediate representation.

Closes #216.

### DIFF
--- a/emanote.cabal
+++ b/emanote.cabal
@@ -1,6 +1,6 @@
 cabal-version:      2.4
 name:               emanote
-version:            0.6.5.7
+version:            0.6.6.0
 license:            AGPL-3.0-only
 copyright:          2021 Sridhar Ratnakumar
 maintainer:         srid@srid.ca
--- a/src/Emanote/View/Template.hs
+++ b/src/Emanote/View/Template.hs
@@ -117,54 +117,58 @@ renderLmlHtml model note = do
       meta = Meta.getEffectiveRouteMetaWith (note ^. MN.noteMeta) r model
       ctx = C.mkTemplateRenderCtx model r meta
       templateName = lookupTemplateName meta
+      -- Force a doctype into the generated HTML. This should probably be done by Heist.
+      -- Issue tracker: https://github.com/srid/emanote/issues/216
+      withDoctype = ("<!DOCTYPE html>\n" <>)
       withLoadingMessage =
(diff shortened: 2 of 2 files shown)


srid:
Thanks! Could you refactor this to

- create a function `withDoctypeForced` (analogous to `withLoadingMessage`) and
- add a comment describing what this workaround does (by linking to #216)

As well as please increment the version in [the cabal file](https://github.com/srid/emanote/blob/master/emanote.cabal), to `0.6.6.0`.

srid:
I have a suggestion to minimize diff, but since I don't have permissions to push to this PR directly I've opened a PR on your repo: https://github.com/ners/emanote/pull/1

If you could merge it, we can merge this one.

ners:
Thanks for the suggestion. I'm sorry you didn't have permissions to edit, maybe I should've enabled something on the PR.

srid:
No worries; it is this option when you create the PR: 

image

srid:
@ners I just came across your neat looking site: https://ners.ch/hello-zettel

Do you think we should add[^1] KaTeX to Emanote and recommend that [in the docs](https://emanote.srid.ca/tips/js/math)? 

[^1]: Alongside mathjax here: https://github.com/srid/emanote/blob/9190d077f317db7dee4f3ec48b44a16a61337192/default/index.yaml#L99-L109

PR #290, opened 2022-05-30
nix: Use `haskell-flake` (a `flake-parts` module)

srid:
https://github.com/srid/haskell-flake


srid:
The use of `overrides` triggers a `error: infinite recursion encountered` Nix error. Full trace at https://gist.github.com/srid/ba6f35c8e65688860f377d9b7a4f7766

image

EDIT: This got fixed in https://github.com/srid/haskell-flake/issues/4

srid:
Garnix's macOS fails with https://github.com/garnix-io/issues/issues/17 - but let's not block the merge.

PR #302, opened 2022-06-12
Allow uptree nodes to be configured

fieldstrength:
Hey there. We talked about this on Matrix a while back. 

Ive been happily using Emanote with a fork that only disabled this implicit node. So to get back on the main release, the presence needs to be configurable. This patch seems to accomplish that as intended.

This does raise the question about whether the special case for the root node is needed. If I were designing around only my preferred usage I would probably set the default value to False instead of True, and remove the special treatment of the root. But since the current behavior corresponds to your preference, I preserved that here. But it's worth noting that it's a possibility in the design space to skip that special treatment.

Feel free to make any branding suggestions. And I'd be happy to add something to documentation, if you have any thoughts on where I should put it (`guide.md`? `tips.md`?)

### DIFF
--- a/src/Emanote/Model/Graph.hs
+++ b/src/Emanote/Model/Graph.hs
@@ -33,16 +33,20 @@ modelFolgezettelAncestorTree r0 model =
           folgezettelFrontlinks =
             frontlinkRels r model
               & mapMaybe (lookupWikiLink <=< selectReverseFolgezettel . (^. Rel.relTo))
+          folgezettelFolder =
+            -- Folders are automatically made a folgezettel
+            maybeToList
+              ( do
+                  lmlR <- leftToMaybe (R.modelRouteCase r)
+                  guard $ lookupRouteMeta True ("emanote" :| ["folder-folgezettel"]) lmlR model
+                  parentLmlRoute lmlR
+              )
           folgezettelParents =
-            folgezettelBacklinks
-              <> folgezettelFrontlinks
-              -- Folders are automatically made a folgezettel
-              <> maybeToList
-                ( do
-                    lmlR <- leftToMaybe (R.modelRouteCase r)
-                    guard $ lookupRouteMeta True ("emanote" :| ["folder-folgezettel"]) lmlR model
-                    parentLmlRoute lmlR
-                )
+            mconcat
+              [ folgezettelBacklinks,
+                folgezettelFrontlinks,
+                folgezettelFolder
+              ]
       fmap catMaybes . forM folgezettelParents $ \parentR -> do
         let parentModelR = R.ModelRoute_LML parentR
         gets (parentModelR `Set.member`) >>= \case
@@ -69,6 +73,7 @@ modelFolgezettelAncestorTree r0 model =
       Rel.RRTFound x -> Just x
       _ -> Nothing
 
+-- | Return the route to parent folder (unless indexRoute is passed).
 parentLmlRoute :: R.LMLRoute -> Maybe R.LMLRoute
 parentLmlRoute r = do
   pr <- do
--- a/src/Emanote/Model/Graph.hs
+++ b/src/Emanote/Model/Graph.hs
@@ -8,6 +8,7 @@ import Data.Tree (Forest, Tree (Node))
 import Emanote.Model.Calendar qualified as Calendar
 import Emanote.Model.Link.Rel qualified as Rel
 import Emanote.Model.Link.Resolve qualified as Resolve
+import Emanote.Model.Meta (lookupRouteMeta)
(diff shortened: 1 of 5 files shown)


srid:
The patch enables excluding _any_ note from the uptree (via `includeInUptree` yaml/frontmatter config). Wasn't the intention to do this for the index route (`index.md` at top-level) only? 

Not that I'm against enabling excluding of any note per see, but I'd like to better understand the use case here.

fieldstrength:
Basically I'd like to turn off any node in the uptree that doesn't come from branching links. With this patch I accomplish that by toggling off the implicit nodes just once at the top level config, but I am relying on the recursive merging behavior of the configuration to also toggle off uptree nodes corresponding to my folders, for notes that are in folders.

So its not just about the root `index.md` node, but you are right that I could get by with a weaker power of just a single top-level setting to toggle all these implicit nodes off at once. That solution would be complete satisfactory to me as well.

(Since you ask about my use case, Im doing a lot of math and physics notes. Back in the Neuron days I got in the habbit of using branching links to represent hierarchies between concepts. So for example stuff like `group -> ring -> field`. Im not sure if this is somehow at odds with the intention of the feature, but I think the results are pretty great :))

This may well be due to my incomplete understanding of the codebase, but: This recursive structure of the configurations and the functions for accessing them seem to encourage me to let this setting be recursively configurable. But you raise a good point: Maybe it's wiser to prefer the simpler configuration surface?

Yet the whole system of accessing the config seems to have this recursive structure, so it's at least not as clear to me how to do it a different way without much bigger changes.

srid:
Oh, I think I had misunderstood what this PR does. 

Notes in Emanote get an **implicit** folgezettel parent that is equivalent to the "folder note". That is, `foo/bar/qux.md` gets an implicit folgezettel parent `foo/bar.md` even if the latter doesn't branch-off link to the former. Uptree just happens to use the folgezettel graph directly.

You want to turn off this implicit (folgezettel) relationship, by setting `includeInUptree` to False. That actually makes sense, and yea - it should be a route-level config so you can set it for the entire sub-tree and forget about it.

src/Emanote/Model/Graph.hs
```diff
@@ -36,7 +37,7 @@ modelFolgezettelAncestorTree r0 model =
             folgezettelBacklinks
               <> folgezettelFrontlinks
               -- Folders are automatically made a folgezettel
-              <> maybeToList (parentLmlRoute =<< leftToMaybe (R.modelRouteCase r))
+              <> maybeToList (parentLmlRoute model =<< leftToMaybe (R.modelRouteCase r))
```
srid: The signature (or behaviour) of `parentLmlRoute` shouldn't have to change here---it is good programming practice to keep the scope of a function as small and narrow as possible. Instead, you want to do the check at the call site. Something like:

```haskell
<> maybeToList (do 
       lmlR <- leftToMaybe (R.modelRouteCase r)
       guard $ lookupRouteMeta True ("emanote" :| ["folder-folgezettel"]) lmlR model
       parentLmlRoute model lmlR)
```

srid: Also, in addition to renaming this key, I'd put it in `./default/index.yaml` which also acts as 'documentation' for available configuration:

```yaml
emanote:
  # Whether to automatically treat folder notes as a folgezettel parent of its contents
  folder-folgezettel: true
```

srid (changes requested):
@fieldstrength Generally looks good to me. Can you also merge `master` and increase the .cabal file version from `0.6.12.0` to `0.6.13.0`?

srid:
> And I'd be happy to add something to documentation, if you have any thoughts on where I should put it (`guide.md`? `tips.md`?)

This would be great. You can put it in `guide/uptree.md`.

fieldstrength:
Thanks for the feedback, and working with me on this! 

I adopted your requests/suggestions, and took a stab at a simple guide page for the uptree.

srid (approved):
LGTM. I'll test and merge if everything's good.

fieldstrength:
Hey I just pushed a small tweak to the doc page before I saw your comment:

- Moved it to `docs/guide/uptree.md` to align with `docs/guide/neuron.md`
- Changed the boolean flag in the doc to `false` to match the text instructions
- Removed an extra blank line
- Gave the note the more descriptive title "Uplink tree"

Thanks again for your efforts!

srid:
@fieldstrength I don't know why we went with this particular semantic, but I'm *inverting* the meaning of `folder-folgezettel` in #483. This setting will now apply for children not parents. I think that makes more sense actually, as you can turn on/off folgezettel for the top-level folder en masse.

EDIT: Let me know if I missed anything.

PR #306, opened 2022-06-18
Add `default.nix`

bolt12:
This PR adds a default.nix so that one can still use this repo from a legacy nix system.

More info here: https://nixos.wiki/wiki/Flakes#Using_flakes_project_from_a_legacy_Nix

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,10 @@
+(import
+  (
+    fetchTarball {
+      url = "https://github.com/edolstra/flake-compat/archive/12c64ca55c1014cdc1b16ed5a804aa8576601ff2.tar.gz";
+      sha256 = "0jm6nzb83wa6ai17ly9fzpqc40wg1viib8klq8lby54agpl213w5";
+    }
+  )
+  {
+    src = ./.;
+  }).defaultNix


flake.nix
```diff
@@ -20,8 +20,12 @@
     ixset-typed.flake = false;
     tailwind-haskell.url = "github:srid/tailwind-haskell/master";
     tailwind-haskell.inputs.nixpkgs.follows = "nixpkgs";
+    flake-compat = {
+      url = "github:edolstra/flake-compat";
+      flake = false;
+    };
```
srid: We don't need to add the `flake-compat` input here, as it can be referenced directly from `default.nix`. See https://github.com/srid/emanote/blob/master/shell.nix

srid (commented):
@bolt12 `flake-compat` was removed explicitly, because the official installation instruction now uses `nix profile install ...`.

https://github.com/srid/emanote/commit/d8888273d66316acb001aa3580f6372bc3e3dbc2

It sounds like you are still on the legacy Nix install (ie., not using flakes yet), so a default.nix is necessary.

I'd prefer to keep Nix files to minimum, but if it helps people on legacy Nix, let's add it back (as long as that extra flake input is not added back).

flake.nix
bolt12: Thank you for your reply! And I apologize for adding the files back. I don't mind having my own fork with this patch! I don't think my decision to keep with legacy nix should be imposed in the repository!

srid: Oh, no worries. I'm still open to adding the old `default.nix` back (but without proxying to the flake input because, as the current shell.nix shows, we don't need that).

Adding one small .nix file to support people on legacy (pre-flakes) Nix is a reasonable tradeoff.

bolt12: I see, thanks I'll edit then!

PR #310, opened 2022-06-24
ema: Use RouteEncoder from siteOutput, not siteInput

srid:


srid:
This PR in its current state breaks Emanote:

image

This is because `fixStaticUrl` in Patch.hs (evaluated during `siteInput`) uses Ema's `encodeRoute` passing the then-non-existant RouteEncoder: https://github.com/srid/emanote/blob/147528d9df81b881214652ce0cefec0b3d52965e/src/Emanote/Source/Patch.hs#L155

src/Emanote/Model/Type.hs
```diff
@@ -78,6 +78,10 @@ modelReadyForView =
 inLiveServer :: Model -> Bool
 inLiveServer = Ema.CLI.isLiveServer . _modelEmaCLIAction
 
+modelRouteEncoderMust :: HasCallStack => Model -> RouteEncoder Model SiteRoute
+modelRouteEncoderMust model =
+  fromMaybe (error "RouteEncoder not initialized") $ model ^. modelRouteEncoder
```
srid: Also not a fan of this hack.

lucasvreis: Perhaps a way without hacks is to wrap the rendering functions in a `Reader (RouteEncoder Model SiteRoute)` monad? Then you can get the encoder at the start and don't bother with parameters. You would have to change more lines, though.

srid:
Working around that one ... another place where URL is required is the 'postRun' stage, which takes the initial model returned by `runSiteWithCli`.

image

https://github.com/srid/emanote/runs/7063245975

`Export.modelRels` throws that error:

https://github.com/srid/emanote/blob/31076442976cd95c88cb345c5cf36e1559ed11b0/src/Emanote.hs#L58-L66

src/Emanote/Model/Type.hs
srid: Using `ModelT f` does the trick, at the expense of slight complexity.

srid:
We are not using generic route deriving yet, but that can be done later.

PR #314, opened 2022-07-21
Add baseUrl and prettyUrls options

omasanori:
See https://github.com/EmaApps/emanote/discussions/312#discussioncomment-3191436 for background.

### DIFF
--- a/nix/emanote.nix
+++ b/nix/emanote.nix
@@ -54,6 +54,16 @@ in
                         description = ''Allow broken links in the static site'';
                         default = false;
                       };
+                      baseUrl = mkOption {
+                        type = types.str;
+                        description = ''Base URL for relative links'';
+                        default = "/";
+                      };
+                      prettyUrls = mkOption {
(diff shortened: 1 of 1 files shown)


nix/emanote.nix
```diff
@@ -54,6 +54,16 @@ in
...
+                      baseUrl = mkOption {
+                        type = types.str;
+                        description = ''Base URL for relative links'';
+                        default = "/";
+                      };
+                      urlStrategy = mkOption {
+                        type = types.str;
+                        description = ''Whether links ends with .html (direct) or not (pretty)'';
+                        default = "direct";
+                      };
```
srid: Since there are only two values possible, it would be better to make this a `types.bool`; what do you think?

omasanori: Agreed.

omasanori: Added `prettyUrls` instead of `urlStrategy`.

PR #318, opened 2022-07-23
Fixed so PrismJS and Mermaid can be used together

applejag:
Fixes PrismJS script to work with Mermaid.

Had to use manual syntax highlighting, and in there filter out the mermaid tags. Also had to do some "is hot-reload" checking to make it more reliable when using `emanote run`

I took inspiration from <https://github.com/PrismJS/prism/issues/2068#issuecomment-535289075>, which the `filter-highlight-all` PrismJS plugin was created to simplify anyway.

Also changed some docs:
- Make Mermaid page be structured the same as the Syntax Highlighting page, and duplicated the diagrams so the docs shows the source code as well.
- Changed to only suggest `<script var="js.prism" />` for Syntax Highlighting

Related to discussion #281

## Sidenote

I tried using <https://prismjs.com/plugins/filter-highlight-all/>, which worked fine until the hot-reload kicked in.

My theory is that while running `emanote run` and editing a file that the hot-reload would recreate the `<head>` element and thereby recreate the PrismJS script tag, making it run again.

In the happy path when PrismJS is loaded in first time, it waits until the document has finished loading, and also letting PrismJS plugins to load as well.

But when it ran again after a hot-reload then the document was already ready and it started syntax highlighting before the `filter-highlighting-all` plugin could register its exceptions, so it still would apply syntax highlighting on those poor `<code class="mermaid">` tags.


srid:
@jilleJr Any reason for us (`default/index.yaml`) not to use [the plugin](https://prismjs.com/plugins/filter-highlight-all/) directly? Looks like it may even perform better.

srid:
(n/m - I failed to read your sidenote :-) )

default/index.yaml
```diff
@@ -87,9 +87,25 @@ js:
   # Syntax highlighting using prism.js
   prism: |
     <!-- Prism.js -->
-    <link href="https://cdnjs.cloudflare.com/ajax/libs/prism/1.26.0/themes/prism-tomorrow.min.css" rel="stylesheet" />
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/prism/1.26.0/prism.min.js"></script>
-    <script src="https://cdnjs.cloudflare.com/ajax/libs/prism/1.26.0/plugins/autoloader/prism-autoloader.min.js"></script>
+    <link href="https://cdnjs.cloudflare.com/ajax/libs/prism/1.28.0/themes/prism-tomorrow.min.css" rel="stylesheet" />
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/prism/1.28.0/prism.min.js" data-manual></script>
```
srid, line 92: Could you add a comment explaining why `data-manual` is required?

default/index.yaml
```diff
@@ -87,9 +87,25 @@ js:
...
+      }
+    }
+    window.isHotReload = window.isHotReload !== undefined;
+    if (window.isHotReload) {
+      highlightAllExceptMermaid();
+    } else {
+      window.addEventListener('load', function() {
+        highlightAllExceptMermaid();
+      });
+    }
```
srid: Could you add a comment explaining what `window.isHotReload` does?

docs/tips/js/syntax-highlighting.md
```diff
@@ -5,19 +5,23 @@ In order to enable syntax highlighting, you must use a client-side JavaScript hi
...
 page:
   headHtml: |
-    <link href="https://cdn.jsdelivr.net/npm/prismjs@1.23.0/themes/prism-tomorrow.css" rel="stylesheet" />
-    <script src="https://cdn.jsdelivr.net/combine/npm/prismjs@1.23.0/prism.min.js,npm/prismjs@1.23.0/plugins/autoloader/prism-autoloader.min.js"></script>
-```
-
-Or, using the alias from the default layer's `index.yaml`:
+    <link href="https://cdnjs.cloudflare.com/ajax/libs/prism/1.28.0/themes/prism-tomorrow.min.css" rel="stylesheet" />
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/prism/1.28.0/prism.min.js"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/prism/1.28.0/plugins/autoloader/prism-autoloader.min.js"></script>
```
srid: Any reason we can't use `<snippet var="js.prism" />` here?

default/index.yaml
srid: I'm now less sure of what this is achieving When I tried `bin/run` locally, the old problem (of syntax highlighting not being applied in some cases when switching routes, in live server) remains.

srid (changes requested):
Thanks; this does solve the original problem. However, the code needs comments explaining what it does in various places.

docs/tips/js/syntax-highlighting.md
applejag: Was unsure if we should keep the non-snippet variants.

Should I remove the non-snippet variants for js.prism & js.mermaid docs?

default/index.yaml
applejag: OK I've finally found the cause here. The `morphdom` from Ema is removing the `<script async src=".../mermaid.js">` tags.

When the `highlightElement` is called, the `autoloader` plugin adds an async script tag to the `<body>` and sets a callback on its `onload`.

But here a race condition comes in. I'm only guessing here, but it seems like either:
- (happy path) the `<script>` is loaded and `onload` is called before it's removed from `<body>`,
- or the `<script>` is removed before the browser even starts loading the script resulting in no outcomes (no errors logs, and no syntax highlighting),
- or the browser starts fetching the JS file and at the point the fetch completes the `<script>` tag has already been removed resulting in code error as the autoloader plugin tries to remove the `<script>` tag itself when it no longer exists.

I've tried to combat this, and the code grew. Not my proudest work at all. What do you think?

It's a little glitchy when first opening a page, but it at least resolves into a syntax highlighted page eventually. And there's no glitchyness after a hot reload or after just loading the statically generated site.

default/index.yaml
applejag: Added:

`<!-- Note: Having data-manual attribute so our custom code controls PrismJS, which allows support of Mermaid & Emanote hot-reloads -->`

docs/tips/js/syntax-highlighting.md
applejag: I've changed to only suggest the `<snippet var="..." />` instead

default/index.yaml
srid: Do you think this whole debounce logic is worth upstreaming to Ema (its `multisite` branch)? cf. https://github.com/EmaApps/ema/issues/69

ie., do we envision doing this for anything other than PrismJS?

srid: In particular, perhaps we can abstract this out in a function (in Ema) and call it in one line as:

```javascript
Ema.invoke(1000, highlightAllExceptMermaid);
```

applejag: Adding a debounce function to upstream Ema library just feels like adding it for the sake of it. I'd vote for no as it's quite the simple function.

Anyway that decision is not written in stone. Could extract it first when a need for it appears in multiple places.

Besides, this code is actually me working around Ema's behavior, but perhaps that could be modified instead. Like when using `emanote run` then on a fresh page load it will:

1. run scripts (on regular page load)
2. run scripts when websocket connected
3. load page again (via websockets) and morph DOM
4. run script after DOM morph

I see you have comments in the JavaScript in the sorts of "`We have to reload <script>'s here on initial page load here, so as to make Twind continue to function`", and I don't know the full depth of this, but I think the best option would be if scripts only ran once and morphing the DOM only happens on page changes and not on initial load.

srid:
@jilleJr Could you move the `debounce` stuff to a separate PR since that appears unrelated? Would be easier to review and merge.

EDIT: you can make it on top of this PR's branch to avoid conflicts.

applejag:
> @jilleJr Could you move the `debounce` stuff to a separate PR since that appears unrelated? Would be easier to review and merge.
> 
> EDIT: you can make it on top of this PR's branch to avoid conflicts.

Ehm sure. The debounce logic is a requirement to get this working during `emanote run` though, but if I get you right then I'll simplify this one and only focus on `emanote gen` here

srid:
I believe this PR caused a regression:

https://emanote.srid.ca/tips/js/syntax-highlighting
image

Issue #320, opened 2022-07-23
PrismJS-highlighted TOML table is wrong, due to Tailwind


applejag:
**Describe the bug**
Tailwind styling messes up PrismJS-highlighted TOML code

**To Reproduce**
1. Enable PrismJS
2. Write TOML snippet in note, e.g:

   ````markdown
   ```toml
   [root]
   foo = "bar"
   ```
   ````
3. See invalid formatting:

   ![image](https://user-images.githubusercontent.com/2477952/180623213-07c07a2c-134a-4002-9dff-2c5d6c37a0b3.png)

**Expected behavior**

Should be rendered as:

![image](https://user-images.githubusercontent.com/2477952/180623221-4a561b5d-cca5-49bf-953c-c24e55876ba1.png)

**Additional context**

The PrismJS generated HTML looks like this:

![image](https://user-images.githubusercontent.com/2477952/180623296-ca7ca97f-3db9-4f5f-b546-33f22457b225.png)

With the culprit being the `table` class that Tailwind also defines.

PrismJS defers to overriding the Tailwind styling, as that's what messing this up: https://github.com/PrismJS/prism/issues/2443#issuecomment-799188475

But how do I change the Tailwind class in Emanote?

PR #321, opened 2022-07-23
Added zk docs

applejag:
Tried out using zk on Emanote, but took some tinkering to get it working.

Though'd I'd share my findings

### DIFF
--- a/docs/resources/zk.md
+++ b/docs/resources/zk.md
@@ -15,11 +15,6 @@ taking tool, which makes it useful for [Obsidian](https://obsidian.md/),
 
 ## Configuring repo
 
-```toml
-[root]
-foo = "bar"
-```
-
 Inside your notes root folder (such as your repository root if you're using Git),
 run the `zk init` command:
 
--- a/docs/resources/zk.md
+++ b/docs/resources/zk.md
@@ -82,7 +82,7 @@ when writing `[[wiki links]]` and `#tags`.
 Available plugins:
 
 - [`zk-nvim`](https://github.com/mickael-menu/zk-nvim) for Neovim 0.5+
-- [`zk-vscode`](https://github.com/mickael-menu/zk-vscode) for Visual Studio Code
+- [`zk-vscode`](https://github.com/mickael-menu/zk-vscode) for [[vscode]]
 
 If you're already using an editor plugin such as `coc.nvim`, or otherwise have
 a different editor that already supports LSP (such as SublimeText) then
--- a/docs/resources/zk.md
+++ b/docs/resources/zk.md
@@ -81,7 +81,7 @@ when writing `[[wiki links]]` and `#tags`.
 
 Available plugins:
 
-- [`zk-nvim`](https://github.com/mickael-menu/zk-nvim) for Neovim 0.5+
+- [`zk-nvim`](https://github.com/mickael-menu/zk-nvim) for [[vim|Neovim]] 0.5+
 - [`zk-vscode`](https://github.com/mickael-menu/zk-vscode) for [[vscode]]
 
 If you're already using an editor plugin such as `coc.nvim`, or otherwise have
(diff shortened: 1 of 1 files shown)


docs/resources/zk.md
```diff
@@ -0,0 +1,94 @@
+---
+page:
+  headHtml: |
+    <snippet var="js.highlightjs" />
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.5.1/languages/ini.min.js"></script>
```
srid, line 5: Why not use PrismJS here?

docs/resources/zk.md
```diff
@@ -0,0 +1,94 @@
...
+generic in the sense that it does not lock in to only supporting a certain note
+taking tool, which makes it useful for [Obsidian](https://obsidian.md/),
+[Neuron](https://neuron.zettel.page/), and Emanote notes collections.
+
+## Configuring repo
+
+```toml
+[root]
+foo = "bar"
+```
```
srid: This TOML config is placed here without any context. What is the reader supposed to do with this? Why does it say "foo"? What is the meaning of this config?

docs/resources/zk.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Editor integration
+
+`zk` provides a LSP (Language Server Protocol) server to allow auto-completion
+when writing `[[wiki links]]` and `#tags`.
+
+Available plugins:
+
+- [`zk-nvim`](https://github.com/mickael-menu/zk-nvim) for Neovim 0.5+
+- [`zk-vscode`](https://github.com/mickael-menu/zk-vscode) for Visual Studio Code
```
srid: Link "Visual Studio Code" to the existing VScode page? Likewise, link "Neovim" to `[[vim|Neovim]]`.

srid (changes requested):
Very interesting; I didn't know about zk-vscode ... I should give it a try!

Maybe we can even add `zk` to emanote-template's flake.nix.

docs/resources/zk.md
applejag: Lol I accidentally committed my test code from #320. Nice catch, will remove!

docs/resources/zk.md
applejag: Sure!

docs/resources/zk.md
applejag: Removed

srid:
It would be good to do https://github.com/srid/emanote-template/issues/8 as part of this. I'll try to get to it, unless someone beats me.

applejag:
> Maybe we can even add `zk` to emanote-template's flake.nix.

> It would be good to do [srid/emanote-template#8](https://github.com/srid/emanote-template/issues/8) as part of this. I'll try to get to it, unless someone beats me.

I'll please skip that part. Nix stuff is one of the things I'm still too scared of to try. So don't wait for me to create PR for that :)

docs/resources/zk.md
```diff
@@ -0,0 +1,89 @@
+---
+page:
+  headHtml: |
+    <snippet var="js.highlightjs" />
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.5.1/languages/ini.min.js"></script>
```
srid, line 5: Imagine if we could just `<snippet var="js.highlightjs-lang.ini" />`. 

(Don't have to do this in this PR.)

applejag: Even better if the snippets could accept arbitrary variables, like

```html
<snippet var="js.highlightjs-lang" lang="ini" />

<!-- uses: -->
<script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.5.1/languages/{{ lang }}.min.js"></script>

<!-- results in: -->
<script src="https://cdnjs.cloudflare.com/ajax/libs/highlight.js/11.5.1/languages/ini.min.js"></script>
```

PR #327, opened 2022-07-30
Full-text-search via Stork

### LINKED ISSUE #324: Add stork-based full-text search
The `emanote gen` generated static site should include interactive search, as demo'ed in https://github.com/EmaApps/emanote/issues/102#issuecomment-1198662665

Ideally, we want the live server to support it as well, but that can happen in a subsequent PR.

---

Use the `which` library to include `stork` as part of emanote install and maybe call stork during postRun:

https://github.com/EmaApps/emanote/blob/c28263848a5e34c93bd3b977cc612387dea7f99f/src/Emanote.hs#L64-L65

applejag:
Closes #324 

## Tasks

- ~~[ ] srid: Handle concurrent reads of `/-/stork.st`~~ (live server is not production friendly)
- [x] srid: #336 (may have to note this as known limitation)
- [x] Finalize search input design for both book and note themes
- [x] .tpl files: refactor, reorganize, avoid DRY
- [x] Write documentation under ./guide

## Changes

- By @jilleJr:
  - Added Stork search component, with modal popup inspired by the one found on https://tailwindcss.com/docs
- By @srid:
  - Adds Nix dependency on the Stork Nix package
  - Emanote exports `/-/stork.st` in live- and statically generated site by running `stork build`

## Preview

Demo: <https://jillejr.github.io/emanote/>

image

image

I even did the blurred backdrop thing from tailwindcss' website :P 

For the narrow layout it's using only the :mag: emoji when the sidebar is collapsed.

image

I was not really sure what to do with the neuron-like page, it currently steals a lot of the vertical view-space:

image

<blockquote><details>
<summary>Old version (kept for just preserving history)</summary>

image

It's also sticky: (difficult to show in a screenshot, but try it out on e.g <https://jillejr.github.io/emanote/resources/vim>)

image

Also works on mobile layout:

image

</details></blockquote>

I used that published GitHub Pages to do some manual testing on my phone.

In a separate branch I made small changes to build the index in the CI pipeline and publish it to that above GitHub Pages site.

- Changes: https://github.com/jilleJr/emanote/compare/feature/stork-search...jilleJr:emanote:feature/stork-search-poc


default/templates/layouts/book.tpl
```diff
@@ -71,9 +74,10 @@
             <apply template="components/metadata" />
             <apply template="/templates/hooks/note-end" />
           </main>
+          <apply template="/templates/hooks/after-main" />
```
applejag: I see some `<apply template>` have `/` prefix, while some dont. Is there a major difference?

docs/templates/hooks/more-head.tpl
```diff
@@ -0,0 +1,68 @@
+<!-- Stork search styling, scripts added in after-body.tpl -->
+<link rel="stylesheet" href="https://files.stork-search.net/releases/v1.5.0/basic.css" />
+<style>
```
applejag: Maybe more suiting to have this as a snippet variable instead?

applejag:
`<rant>`

One thing to consider here is that this has a lot of impact on how Emanote sites are used, and I predict they will also have a great impact on how they are written as well.

On the negative side, this search feature could:

- render tags useless.
- remove importance on creating proper folgezettel links to connect your knowledge.

This isn't just an added positive feature, it also dulls down the other features of Emanote.

On the other hand, the positives are things like:

- user can search for related topics if they've forgotten the main subject, and then navigate via folgezettel links to the one they sought.
- speeds up usage when you know ahead-of-time what subject you're searching for, as you don't have to navigate through collapsed folders in the sidebar.

---

Do we want to counteract the negatives somehow? Or document the pros/cons of using search vs tags & uplink trees?

This could actually be a "killer feature" that makes users jump to writing docs in Emanote instead of Hugo or Jekyll. Most programs are after all only as strong as their search feature.

`</rant>`

default/templates/layouts/book.tpl
srid: IIUC - one is relative, and the other is absolute (from templates/ directory).

srid:
Note to self: in addition to the UI changes in this PR, [stork index generation](https://github.com/EmaApps/emanote/compare/master...jilleJr:emanote:feature/stork-search-poc#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR100-R106) should be done in postRun.

```
          STORK_TOML=./-/stork.toml
          STORK_INDEX=./-/stork.st
          echo '[input]' > $STORK_TOML
          echo 'files = [' >> $STORK_TOML
          grep '<title>' -A1 -r --exclude-dir '-' --include '*.html' | grep html- | sed 's/\(.*.html\)-  *\(.*\)/  {path = "\1", url="\1", title = """\2"""},/' | tee -a $STORK_TOML
          echo ']' >> $STORK_TOML
          /opt/stork build --input $STORK_TOML --output $STORK_INDEX
```

We want to use emanote's model internally rather than parsing HTML files. Also add `stork` as nix dependency and use staticWhich.

srid:
@jilleJr Thanks! I'll hook the rest of the machinery up, and see if we can have `emanote gen` automate it all.

applejag:
> Thanks! I'll hook the rest of the machinery up, and see if we can have `emanote gen` automate it all.

Thank you so much! I can create docs for this later (after this PR is merged, when Emanote's behavior on this is decided)

default/templates/layouts/note.tpl
```diff
@@ -6,12 +6,13 @@
 
 <apply template="base">
   <bind tag="head-main">
-    <link rel="stylesheet" href="${ema:emanoteStaticLayerUrl}/inverted-tree.css" />
+    <link rel="stylesheet" href="/${ema:emanoteStaticLayerUrl}/inverted-tree.css" />
```
applejag: Oops, didn't mean to commit that.

```suggestion
    <link rel="stylesheet" href="${ema:emanoteStaticLayerUrl}/inverted-tree.css" />
```

docs/templates/hooks/more-head.tpl
applejag: I've moved it into being a snippet. I felt it was easier to manage when handling supporting multiple layouts by more easily switch between which script collection to use.

applejag:
Latest changes:
- Added support for neuron-like note layout
- Moved all CSS & JS to snippets
- Moved searchbox to component instead of template hook, and make it visible via CSS snippet

Demo of neuron-like note layout with search box: https://jillejr.github.io/emanote/demo/neuron-layout.html

| image | image
| --- | --- |

applejag:
On second thought, this is now how I envision to enable search, where it's a 2-step process:

1. Add command flag or YAML config value to make emanote create the `/-/stork.st` index file
2. Add the `<snippet var="js.stork-search" />` to your `page.headHtml` in `index.yaml`

srid:
@jilleJr Give 12cf3a1 a try. The stork index is built behind the scenes in both live server and generate mode.

Note to self:

- [ ] Performance must be bad; so we want to cache the index in memory instead of rebuilding it on every request.
- [ ] Multiple layer (`-L`) scenario probably is broken here. Need to fix that.

srid:
Design-wise, I'm not sure about the current style. Can we do something like the search UX in https://tailwindcss.com/ ?

Instead of a center search bar, we could have a small search icon somewhere not-too-prominent-but-yet-noticeable - but retain the Tailwind site's Cmd+K/Ctrl+k keybinding, along with the search results appearing on the center screen. The [flat](https://stork-search.net/themes) theme is not bad.

Also for some strange reason the search bar is not even rendering on certain pages (math and mermaid).

applejag:
> Design-wise, I'm not sure about the current style. Can we do something like the search UX in https://tailwindcss.com/ ?
> 
> Instead of a center search bar, we could have a small search icon somewhere not-too-prominent-but-yet-noticeable - but retain the Tailwind site's Cmd+K/Ctrl+k keybinding, along with the search results appearing on the center screen. The [flat](https://stork-search.net/themes) theme is not bad.

Sure! I'll attempt something! I'll try out having a modal and a discrete search icon.

> Also for some strange reason the search bar is not even rendering on certain pages (math and mermaid).

Ah yes, my bad. Those pages overwrite the `page.headHtml`. Fixed!

srid:
> Ah yes, my bad. Those pages overwrite the page.headHtml. Fixed!

Users shouldn't have to do this (dc82292bade41cf85b11d7758ad3888c4991a531), though. I think search should be a built-in feature, not just something users toggle on/off on individual pages. Maybe it is best to put them in `.tpl` files. I'll think about this more later.

applejag:
> Users shouldn't have to do this ([dc82292](https://github.com/EmaApps/emanote/commit/dc82292bade41cf85b11d7758ad3888c4991a531)), though. I think search should be a built-in feature, not just something users toggle on/off on individual pages. Maybe it is best to put them in `.tpl` files. I'll think about this more later.

Sure, I've changed it back so it's only added via templates now.

But I can't run it locally, it fails to compile. I probably did something wrong when resolving the merge conflicts. @srid could you take a look?

```
src/Emanote.hs:53:23: error:
    • Couldn't match type ‘Ema.Asset.Asset LByteString’
                     with ‘Data.ByteString.Lazy.Internal.ByteString’
      Expected: m (SiteOutput SiteRoute)
        Actual: m (Ema.Asset.Asset (Ema.Asset.Asset LByteString))
    • In the expression: pure $ View.emanoteSiteOutput rp m r
      In an equation for ‘siteOutput’:
          siteOutput rp m r = pure $ View.emanoteSiteOutput rp m r
      In the instance declaration for ‘EmaSite SiteRoute’
   |
53 |   siteOutput rp m r = pure $ View.emanoteSiteOutput rp m r
   |                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

I'm waiting with making the design changes while I can't compile the program. Sorry, I'm pretty lost when it comes to Haskell stuff

srid:
@jilleJr I've fixed it in 460a1f2

applejag:
@srid What about this now?
https://jillejr.github.io/emanote/

image

image

I even did the blurred backdrop thing from tailwindcss' website :P 

For the narrow layout it's using only the :mag: emoji when the sidebar is collapsed.

image


I was not really sure what to do with the neuron-like page, it currently steals a lot of the vertical view-space:

image

Thoughts?

Bipodos:
From an early adopter who knows very little about programming: I managed to install Emanote via 'nix profile --impure install github:jilleJr/emanote/feature/stork-search' on my Mac. Both gen and run create a website with search enabled, but nothing can be searched for. I get messages
[Warn#emanote] Generating search index using Stork (this may be expensive)
[Info#emanote] Done generating Stork index
but stork.st is 0 bytes long in static version (and I suppose is also empty in the run mode). 

When installing Emanote with stork on Mac, I need to use the option --impure. Without it, I get the message:
 Package ‘stork-1.4.2’ in /nix/store/a885zpv9ys2p2x7qnzqvxlsy321mclip-source/pkgs/applications/misc/stork/default.nix:29 is marked as broken, refusing to evaluate.
Is stork broken for Mac? Or should I force using cache.garnix.io (I enabled it, per instructions, but still get the message 
"ignoring untrusted substituter 'https://cache.garnix.io'"

applejag:
> From an early adopter who knows very little about programming: I managed to install Emanote via 'nix profile --impure install github:jilleJr/emanote/feature/stork-search' on my Mac. Both gen and run create a website with search enabled, but nothing can be searched for. I get messages [Warn#emanote] Generating search index using Stork (this may be expensive) [Info#emanote] Done generating Stork index but stork.st is 0 bytes long in static version (and I suppose is also empty in the run mode).

Interesting! And concerning!

I'm unable to reproduce this, as it seems to work just fine at least on the Emanote docs.

@Bipodos  Which repo have you tested this on? Do you get the same result when checking out this branch and running the `bin/run` script?

Sample working (for me/on my machine):

```console
(inside cloned https://github.com/jilleJr/emanote, branch feature/stork-search)
$ mkdir -p result
$ nix run github:jilleJr/emanote/feature/stork-search -- gen result -L docs --allow-broken-links

$ du -h result/-/stork.st
396K	result/-/stork.st
(^ not 0 bytes)

(also, can perform searches on it:)
$ stork search --index result/-/stork.st --query foo
{
  "results": [
    {
      "entry": {
        "url": "demo/markdown.html",
        "title": "Extended Markdown \\U0000270d\\U0000fe0f",
        "fields": {}
      },
      "excerpts": [
        {
          "text": "footnotes. You may also reuse[^1] footnotes. [^1]: First footnote example [^2]: Second footnote example. Footnotes within [^1] footnotes are not handled. Task",
          "highlight_ranges": [
...
```

---

Sidenote:
However when playing around I encountered a different error when running stork on my own notes page (https://github.com/jilleJr/notes) where the resulting `stork.st` seems to have some issue (or there's some issue in Stork):

```console
(inside cloned https://github.com/jilleJr/notes)
$ mkdir -p result
$ nix run github:jilleJr/emanote/feature/stork-search -- gen result -L content
$ du -h result/-/stork.st
748K	result/-/stork.st
(^ also not 0 bytes)

(but panics when performing searches on it:)
$ stork search --index result/-/stork.st --query foo
thread 'main' panicked at 'split_to out of bounds: 764782 <= 763741', /home/kalle/.cargo/registry/src/github.com-1ecc6299db9ec823/bytes-1.1.0/src/bytes.rs:402:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I'll bring up the above panicked error message over at Stork's GitHub repo.

applejag:
> When installing Emanote with stork on Mac, I need to use the option --impure. Without it, I get the message:
> /..snip../
> Is stork broken for Mac?

Stork is explicitly supported on Mac https://stork-search.net/docs/install, though I don't know if there's any Mac-related issues on Stork v1.5.0 or perhaps with the Nix-packaged version.

I don't have a Mac so I can't verify this. @srid do you?

Bipodos:
1. I checked out `github:jilleJr/emanote/feature/stork-search`
2. I run Emanote locally, after installing in line with instructions at https://emanote.srid.ca/start/install
3. Stork is installed by Emanote in the nix profile, but by default, Emanote asks for 1.4.2. At [Nixos stork package site](https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=stork) one can see that 1.5.0 is treated as "unstable".  In [Nix package definition for 1.5.0](https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/applications/misc/stork/default.nix#L33)  it is described as 
```
    # TODO: Remove once nixpkgs uses macOS SDK 10.14+ for x86_64-darwin
    # Undefined symbols for architecture x86_64: "_SecTrustEvaluateWithError"
    broken = stdenv.isDarwin && stdenv.isx86_64;
```
I have an Intel MacBook Pro, this is the architecture described as broken. Still, even 1.4.2 is treated as broken
Anyway, the installation of stork done via nix profile is visible only from the profile. I cannot run stork myself. Should one install stork separately?   
4. I did it (just downloading a binary from stork, making executable, putting in a directory with path to it). It can be called manually, but Emanote installed as package produces the same 0 byte index. I shall try running directly from repository.
5. Anyway, can having lots of Greek in the vault something to do with it? Is stork unicode-aware? I am trying to move a project dictionary online

applejag:
OK just to try confirm that Stork may be working, could you do the following?

1. Create file `stork.toml`
   ```toml
   [input]
   files = [
     {path = "example.txt", url = "0001", title = "Example"},
   ]
   ```
2. Create file `example.txt`
   ```plaintext
   Lorem ipsum
   Alphabet: Α α, Β β, Γ γ, Δ δ, Ε ε, Ζ ζ, Η η, Θ θ, Ι ι, Κ κ, Λ λ, Μ μ, Ν ν, Ξ ξ, Ο ο, Π π, Ρ ρ, Σ σ/ς, Τ τ, Υ υ, Φ φ, Χ χ, Ψ ψ, Ω ω.
   ```
3. Build the index via your downloaded Stork CLI (that you mentioned in your step 4.)
   ```console
   $ ./stork build -i stork.toml -o stork.st
   Success: Index built successfully, wrote 1,171 bytes.
   Index stats:
     - 1 entries
     - 42 search terms
     - 1,171 bytes per entry
     - 27 bytes per search term

   $ du -h stork.st
   4.0K	stork.st
   ```
4. Try building via the Stork Nix package:

   ```console
   $ nix run "nixpkgs#stork" -- build -i stork.toml -o stork.st
   Success: Index built successfully, wrote 1,171 bytes.
   Index stats:
     - 1 entries
     - 42 search terms
     - 1,171 bytes per entry
     - 27 bytes per search term

   $ du -h stork.st
   4.0K	stork.st
   ```

5. If both above succeed, just a sanity check that search also works:
   ```console
   $ ./stork search -i stork.st -q alph
   {
     "results": [
       {
         "entry": {
           "url": "0001",
           "title": "Example",
           "fields": {}
         },
         "excerpts": [
           {
             "text": "Lorem ipsum Alphabet: Α α, Β β, Γ γ, Δ",
   ...

   $ nix run "nixpkgs#stork" -- search -i stork.st -q alph
   (same output)
   ```

Bipodos:
After adding stork independently from nix, and after running Emanote directly from repository, so as @jilleJr does, I got the same message about broken stork
```
a) To temporarily allow broken packages, you can use an environment variable
          for a single invocation of the nix tools.

            $ export NIXPKGS_ALLOW_BROKEN=1

        Note: For `nix shell`, `nix build`, `nix develop` or any other Nix 2.4+
        (Flake) command, `--impure` must be passed in order to read this
        environment variable.

       b) For `nixos-rebuild` you can set
         { nixpkgs.config.allowBroken = true; }
       in configuration.nix to override this.

       c) For `nix-env`, `nix-build`, `nix-shell` or any other Nix command you can add
         { allowBroken = true; }
       to ~/.config/nixpkgs/config.nix.
(use '--show-trace' to show detailed location information)
```
Works with `-- impure` after `run`
But result is identical
``` 
du -h  stork.st                            
  0B	stork.st
```
Nb. I cannot cd to directory named  "-" on Mac via `cd -` (goes to former  directory of the terminal instead). I cannot read the file from an upper directory. Only showing full path in du works. If you are using abbreviated relative paths, may be some commands cannot find the file?

applejag:
> Nb. I cannot cd to directory named "-" on Mac via `cd -` (goes to former directory of the terminal instead)

Small tip: you can do `cd ./-` or `cd -/` instead. This trick works on most CLI programs that has `-` overridden, e.g `wget -O -` (targets STDOUT) vs `wget -O ./-` (targets file name `-`)

Bipodos:
@jilleJr, working on your instructions
1.2. done
3.:
```
Stork % stork build -i stork.toml -o stork.st
Success: Index built successfully, wrote 1,171 bytes.
Index stats:
  - 1 entries
  - 42 search terms
  - 1,171 bytes per entry
  - 27 bytes per search term
% du -h stork.st
4,0K	stork.st
```
This runs on a binary downloaded from [stork github](https://github.com/jameslittle230/stork/releases/download/v1.5.0/stork-macos-10-15) and put in /opt/local/bin, a location to which user can point PATH (/usr/bin/ is off limits in Mac now).
4.: 
``` 
 % nix run "nixpkgs#stork" -- build -i stork.toml -o stork.st
error: Package ‘stork-1.5.0’ in /nix/store/0c10w4am1xhxjkrmnd8d6rbksvm68311-source/pkgs/applications/misc/stork/default.nix:29 is marked as broken, refusing to evaluate.
```
And a message as in a [comment above](https://github.com/EmaApps/emanote/pull/327#issuecomment-1214382263)

I do `export NIXPKGS_ALLOW_BROKEN=1` and run `nix run --impure "nixpkgs#stork" -- build -i stork.toml -o stork.st`
Takes some time, because does lot of download and compiles in `/private/tmp/nix-build-stork-1.5.0.drv-0`
Worked:
```
Success: Index built successfully, wrote 1,171 bytes.
Index stats:
  - 1 entries
  - 42 search terms
  - 1,171 bytes per entry
  - 27 bytes per search term
````
5.: checking search from binary in /opt/
````
 % stork search -i stork.st -q alph
{
  "results": [
    {
      "entry": {
        "url": "0001",
        "title": "Example",
        "fields": {}
      },
      "excerpts": [
        {
          "text": "Lorem ipsum Alphabet: Α α, Β β, Γ γ, Δ",
          "highlight_ranges": [
            {
              "beginning": 12,
              "end": 21
            }
          ],
          "score": 123,
          "internal_annotations": [],
          "fields": {}
        }
      ],
      "title_highlight_ranges": [],
      "score": 246
    }
  ],
  "total_hit_count": 1,
  "url_prefix": ""
}
```
Via nix:
```
nix run "nixpkgs#stork" -- search -i stork.st -q alph
error: Package ‘stork-1.5.0’ in /nix/store/0c10w4am1xhxjkrmnd8d6rbksvm68311-source/pkgs/applications/misc/stork/default.nix:29 is marked as broken, refusing to evaluate.
``` etc
with `--impure`

```
nix run --impure "nixpkgs#stork" -- search -i stork.st -q alph
{
  "results": [
    {
      "entry": {
        "url": "0001",
        "title": "Example",
        "fields": {}
      },
      "excerpts": [
        {
          "text": "Lorem ipsum Alphabet: Α α, Β β, Γ γ, Δ",
          "highlight_ranges": [
            {
              "beginning": 12,
              "end": 21
            }
          ],
          "score": 123,
          "internal_annotations": [],
          "fields": {}
        }
      ],
      "title_highlight_ranges": [],
      "score": 246
    }
  ],
  "total_hit_count": 1,
  "url_prefix": ""
}
```
Took a slice of second longer than the former approach.
So the problem seems to be elsewhere.

**Comment by srid** [comment]
This PR works on my M1 mac, but I don't have an intel mac to test (and we don't use it in CI; #335). In any case, I just upgraded stork to 1.5.0 in this PR. Not sure if that will help or not, @Bipodos 

@jilleJr I like your recent design changes. I will have to think about where to put the search box as it is a bit jarring on the `note` layout definitely; even the `book` layout's position is not ideal. By the way, on mac `Cmd+K` should also work. I'm yet to review the `.tpl` files themselves.

**Comment by applejag** [comment]
Btw I added a workaround in b308b8db  for the issue I experienced earlier (https://github.com/EmaApps/emanote/pull/327#issuecomment-1214382184), basically changing:

```bash
stork build --output -
```
to
```bash
stork build --output /dev/stdout
```

---

> By the way, on mac Cmd+K should also work.

On it!

srid:
@Bipodos Try the latest PR (after 2b8ff25); it should work now on intel mac.

applejag:
Commit 0d96667 containing the added Cmd+K and Esc keybindings have been deployed to https://jillejr.github.io/emanote/

Bipodos:
I recompiled Emanote after changes through `nix profile install github:jilleJr/emanote:feature/stork-search`, also did ` nix run github:jilleJr/emanote/feature/stork-search -- gen "/Users/... -L "/Users/.../content/"`
Do not have to run with `--impure` s
Zrzut ekranu 2022-08-14 o 18 27 13
witch anymore. Cmd-K works, Otherwise, index `stork.st` is 0 bytes long. Obviously, nothing can be found.

May be there is something wrong with the material to be indexed (not just Greek, but its ancient variant) or with index.yaml
I will give the developers access to a copy of the source vault. If my wife lets me using her M1, may be I see whether this this processor architecture problem. Anyway, I do not suppose I could mess things up to this level that the site is created, just with 0 bytes index.

srid:
Emojis are broken in the search results. The `\U000...` part.

image

applejag:
> Emojis are broken in the search results:

Yep I noticed that too. Also some Markdown formatting is leaking through, such as the pipes in tables and the wiki-links.

Would it be possible/feasible to send the HTML-rendered pages to Stork instead of the markdown?

srid:
> Would it be possible/feasible to send the HTML-rendered pages to Stork instead of the markdown?

Not in live server (and we want Stork to work in live server the same way it does in statically generated site, as this PR currently demonstrates), because of incremental build (HTML is rendered only for the requested page, on demand).

This is why Emanote has to use Tailwind 2.x in live server, but Tailwind 3.0 in generated mode. 3.0 has no CDN, and forces the use of JIT compiler.

applejag:
@Bipodos Sad to say, but your project mostly just works fine _on my machine_.

I faced some issues, but resulting in a 0 byte stork.st file was never one of them.

The issues I faced was:
- I had to set `LANG=C.UTF-8` & `LC_ALL=C.UTF-8` when running Emanote, otherwise it just complained about "invalid character" in the filenames.
- Emanote was complaining HEAVILY about broken links ("Found 192 broken links!") for files that I can find when doing regular `find` for the filename.
- The index doesn't seem to contain all files. Probably related to above.

The missing files is their own topic that deserves a separate GitHub issue, but from your project I could at least search in the file it did include, e.g here where "termin" could be found in 3 different files:

```console
$ stork search --index result/-/stork.st --query termin
{
  "results": [
    /..snip../
  ],
  "total_hit_count": 3,
  "url_prefix": ""
}
```

For reference, I'm running on Intel, but on GNU/Linux.

applejag:
> Emojis are broken in the search results. The `\U000...` part.

@srid I found the cause. I made Emanote write the TOML file it pipes to stork to a file, and in it we find:

```toml
[[input.files]]
  path = "./docs/demo/markdown.md"
  title = "Extended Markdown \\U0000270d\\U0000fe0f"
  url = "demo/markdown.html"
```

It's escaping the unicode-syntax. It should only be 1 backslash there.

image

I.e, it seems to be somewhere in the haskell code that's causing this

Bipodos:
Thank you, @jilleJr, for checking up my project. This seems, thus, to be an issue on Intel Macs. I shall see on M1 (may be only in a few days). UTF-8 is default for Mac, thus I had no "invalid characters", and missing links will disappear when I end  moving the project from OpenProject to markdown/Obsidian/Emanote. I shall let you know once compile Emanote on a M1 Mac.

applejag:
> This seems, thus, to be an issue on Intel Macs. 

This is though a sample size of 1, which means there are a lot more variables at play here. I know some coworkers has Macs, maybe intel ones. I'll ask if any of them are willing to test this out as well.

Bipodos:
I suppose the issue is about Intel Macs because I have an extremely limited installation of Nix, only Emanote in it, and I have very little set in respect of variables etc. But we shall see. Thank you once more!

srid:
@Bipodos I tried https://github.com/Bipodos/Plotinus on my M1 mac, but under rosetta emulation (passing `--option system x86_64-linux` to nix) - and I cannot reproduce your issue.

image

srid:
Note: if anybody wants to test this feature, run the following after cd'ing to your notebook:

```
nix run github:jilleJr/emanote/feature/stork-search
```

default/templates/components/stork/stork-search-css.tpl
```diff
@@ -0,0 +1,18 @@
...
+
+	.stork-overflow-hidden-important {
+		overflow: hidden !important;
+	}
+
+	.stork-wrapper-flat .stork-close-button svg {
+		/* small bugfix in Stork's stock styling */
+		margin: auto;
+	}
+</style>
```
srid: @jilleJr Why do we need this file? Can't we do the styling inline (in `stork-search.tpl`, etc.) using Tailwind classes?

default/templates/components/stork/stork-search-js.tpl
```diff
@@ -0,0 +1,45 @@
...
+			window.searchShown = false;
+			function toggleSearch() {
+				document.getElementById('stork-search-container').classList.toggle('hidden');
+				window.searchShown = document.body.classList.toggle('stork-overflow-hidden-important');
+				if (window.searchShown) {
+					document.getElementById('stork-search-input').focus();
+				}
+			}
+
+			function clearSearch() {
```
srid: This is used only in the 'close' button. I think we should remove this function, as well as get rid of the 'close' button because it is superfluous (the user can always close the modal by clicking elsewhere).

default/templates/components/stork/stork-search-js.tpl
```diff
@@ -0,0 +1,45 @@
+<script src="https://files.stork-search.net/releases/v1.5.0/stork.js"></script>
+<ema:metadata>
+	<with var="template">
+		<script data-emanote-base-url="${value:baseUrl}">
+			window.searchShown = false;
+			function toggleSearch() {
```
srid: I'd like to see functions like `toggleSearch` not be made top-level; but rather inside a set, like `Emanote.toggleSearch`. Because then it won't have to conflict with user-defined top-level functions.

default/templates/components/stork/stork-search-css.tpl
srid: If there is no reason for it to be having separate, I'd also inline `stork-search-js.tpl` in `stork-search.tpl`.

default/templates/components/stork/stork-search-js.tpl
applejag: Good idea. Will change!

default/templates/components/stork/stork-search-js.tpl
applejag: Sure. Do we want a clear button though? The default behavior from Stork's JS is to add a "close button" that actually just clears the input field.

Change back to a "clear input field" button, or no button at all?

default/templates/components/stork/stork-search-css.tpl
applejag: Sure I can inline them, and I'll try move as much as possible into Tailwind classes. I think the z-index was especially troublesome but I'll give it another go.

default/templates/components/stork/stork-search-js.tpl
srid: I'd remove them all, to keep the UI minimal.

Bipodos:
Zrzut ekranu 2022-08-21 o 11 08 36
> @Bipodos I tried https://github.com/Bipodos/Plotinus on my M1 mac, but under rosetta emulation (passing `--option system x86_64-linux` to nix) - and I cannot reproduce your issue.
> 
> image

After having updated nix, the stork search now works on my Intel Mac. 
Last week, I also used the most recent version of nix. After waiting a few days, I gave it a new try. Your updated versions of the resulted in Emanote producing the same 0-byte stork index, until I updated nix. Now index works both in static and run modes.

default/templates/components/stork/stork-search-css.tpl
srid: This can be done in separate PR. Let's get this PR merged right away.

applejag:
@srid Thank you for applying the changes!

PR #328, opened 2022-08-01
Reliable PrismJS rendering in live server (`emanote run`)

applejag:
Based on #318

Adds the "debounce stuff" that was omitted from #318 to make that PR simpler.


srid:
@jilleJr Can this PR be simplified if we implemented your suggestion here on Ema? https://github.com/EmaApps/ema/issues/69#issuecomment-1198463127

EDIT: using https://github.com/EmaApps/ema/pull/127

srid:
For some reason, this stopped working today. Live server syntax highlighting is broken 100% of the time; and I see:

image

applejag:
Big sigh. Yea I've been thinking about this a bit, and this current implementation is quite unstable in nature, especially with the `setTimeout(..., 1000)`, which is a very arbitrary timeout duration that in reality depends on a lot of factors.

Currently I'm thinking maybe it's best to try add debounce on PrismJS autoloader's script creation & loading. I'll goof around with the code a bit more and see what I can cook up.

Won't happen today, as I have other things in focus for tonight.

srid:
As of 0061324 the following events are available: 

image

Feel free to play with it whenever you have time if they help. I'm not sure if they do.

applejag:
> As of [0061324](https://github.com/EmaApps/emanote/commit/0061324aeaa0352a5a0b4bdbf9328ce78478f701) the following events are available:

That's awesome! That'll probably come in really handy here.

> Feel free to play with it whenever you have time if they help. I'm not sure if they do.

I'm currently wondering if I have to resort to [MutationObservers](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver), but that would blow up the size of the script. Keeping it simple is the tricky part.

srid:
Another question to ask: can we make this solution general enough, so it works with other scripts (eg: highlightJS)?

applejag:
> Another question to ask: can we make this solution general enough, so it works with other scripts (eg: highlightJS)?

Don't know, but I'd rather leave that for a future generalization of the problem. I thought HighlightJS didn't have this problem though, right?

srid:
The last time I used HighlightJS, I had to manually pull in language-specific support; is there an automatic way to do that? If so, I'd switch completely to highlightJS.

srid:
Sorry, I had to revert your other PR in https://github.com/EmaApps/emanote/pull/331 ... prismJS was broken after it. We need a comprehensive and thorough solution if we are going to change this already-broken-somewhat-in-live-server PrismJS glue.

Meanwhile, I think we should just recommend highlightJS (despite the inconvenience of explicitly including languages), because it works more reliably on live server.

applejag:
> Sorry, I had to revert your other PR in #331 ... prismJS was broken after it. We need a comprehensive and thorough solution if we are going to change this already-broken-somewhat-in-live-server PrismJS glue.
> 
> Meanwhile, I think we should just recommend highlightJS (despite the inconvenience of explicitly including languages), because it works more reliably on live server.

Yea for sure, and I'm in no way married to the idea of fixing this. If PrismJS is too tedious, which it seems to be, then just advising against it seems reasonable.

Especially with the news that it broke on a statically generated site (emanote.srid.ca) then I'm honestly giving up. Gave it a shot, but oh well, highlightJS is also nice.

PR #341, opened 2022-08-24
Add support for embedded pdf

mnaoumov:
Add support for embedded PDFs in markdown links and wikilinks

Required in #24


srid:
This is a good feature to add; thanks, will review soon.

default/templates/hooks/more-head.tpl
```diff
@@ -34,4 +34,9 @@
   .header-font {
     font-family: 'MavenPro', sans-serif;
   }
+
+  .embedded-pdf {
+    width: 100%;
+    height: 800px;
+  }
```
srid, line 41: You should remove these in favour of inline styles in `embed-pdf.tpl`. Prefer Tailwind classes wherever possible; for example, use `w-full` instead of `width: 100%;`.

srid: You should remove these in favour of inline styles in `embed-pdf.tpl`. Prefer Tailwind classes wherever possible; for example, use `w-full` instead of `width: 100%;`.

src/Emanote/Pandoc/Renderer/Embed.hs
```diff
@@ -82,6 +82,9 @@ embedStaticFileRoute model wl staticFile = do
       | any (`T.isSuffixOf` toText fp) videoExts -> do
         pure . runEmbedTemplate "video" $ do
           "ema:url" ## HI.textSplice url
+      | any (`T.isSuffixOf` toText fp) ["pdf"] -> do
```
srid: This should be simplified to: 

```haskell
".pdf" `T.isSuffixOf` toText fp
```

Note also the dot before "pdf".

src/Heist/Extra/Splices/Pandoc/Render.hs
```diff
@@ -224,10 +224,16 @@ rpInline' ctx@RenderCtx {..} i = case i of
...
+    let lowerUrl = T.toLower url
+    if T.isInfixOf ".pdf?t=" lowerUrl || T.isSuffixOf ".pdf" lowerUrl
+      then do
+        tpl <- HE.lookupHtmlTemplateMust "/templates/filters/embed-pdf"
+        HE.runCustomTemplate tpl ("ema:url" ## HI.textSplice url)
+      else do
+        let attrs =
+              catMaybes [pure ("src", url), guard (not $ T.null tit) >> pure ("title", tit), pure ("alt", WL.plainify is)]
+                <> rpAttr (rewriteClass ctx attr)
+        pure $ one . X.Element "img" attrs $ mempty
```
srid: `src/Heist/Extra` is to be considered a general library; it shouldn't be hardcoded with Emanote-specific stuff (such as the templates).

What is this code doing here anyway? When I revert this change, the PDF demo still works.

srid (changes requested):
@mnaoumov Functionality-wise, it looks good to me.  I've left a few comments to address the code. Do you think there is a way to let the user customize the height instead of hardcoding it?

src/Heist/Extra/Splices/Pandoc/Render.hs
mnaoumov: This code is required for markdown links `book`

srid: Why would we want to embed regular markdown links?

mnaoumov: Without this code, pdf is rendered as `` tag which is not working

mnaoumov: @srid sorry I made a typo. I modified my previous comment, I added the missing exclamation mark

srid: Is linking to pdf using image syntax a common practice? What other editors (Obsidian?) support this non standard practice?

mnaoumov: @srid yes, Obsidian uses this syntax. I didn't use other PKM but I think such syntax the most likely is used for pdfs heavily

srid: Okay, that makes sense: https://help.obsidian.md/How+to/Embed+files#Developer+notes

I think the implementation should then be refactored. Maybe even add a `/templates/filters/embed-image` template so images are not treated in a special manner. I'll do this latter.

mnaoumov: @srid Well, ideally the underline parser distinguishes between `Image` and `Pdf` token, but  as it is not the case now, I consider what I wrote as a hack

srid: > I think the implementation should then be refactored.

@mnaoumov If you'd like to take a stab at doing this, check out the `InlineRef` type (we could add an `InlinePDF` constructor in addition to `InlineImage`), as well as the `urlResolvingSplice` function.

https://github.com/EmaApps/emanote/blob/6433224d6a23577995e9c14fa3bdd75bddf815d8/src/Emanote/Pandoc/Link.hs#L10-L22

https://github.com/EmaApps/emanote/blob/6433224d6a23577995e9c14fa3bdd75bddf815d8/src/Emanote/Pandoc/Renderer/Url.hs#L42-L45

srid: I'll merge this PR, but without the regular image linking support because it will take some time to refactor Emanote to do this properly. It can happen in separate PR.

mnaoumov:
@srid why did you revert the change for markdown link pdf?

Then `` won't render properly

srid:
@mnaoumov Yea, I know. I'm happy to accept the feature itself, but the implementation has to be done differently. See #350

I don't have time to do this right now, so I merged this PR without that syntax.

PR #351, opened 2022-09-23
Allow embedding of PDF, video, etc. using image links

### LINKED ISSUE #350: Support embedding PDF using regular Markdown syntax
ie., `` should also embed the PDF file like it does for images.

[This change](https://github.com/EmaApps/emanote/pull/341/commits/fd885e7eca81d81fb017eb24cd90cb095574785d) (made in #341, but reverted) functionally achieves it, but we want to do it in a simpler manner without overloading `Heist.Extra` with emanote-specific functions. `Heist.Extra` will eventually be made a library of its own; thus, it should not contain code specific to Emanote templates.

srid:
Resolves #350

FYI @mnaoumov https://github.com/EmaApps/emanote/pull/341#issuecomment-1250157729 - this is how it should be done.

### DIFF
--- a/docs/demo/embed.md
+++ b/docs/demo/embed.md
@@ -26,7 +26,7 @@ See https://github.com/srid/emanote/issues/24 for progress.
 
 Embedding image files as, say, `![[disaster-girl.jpg]]` is equivalent to `![](path/to/disaster-girl.jpg))` (this example links to [[disaster-girl.jpg|this image]]).  See also the tip: [[adding-images]].
 
-[![](disaster-girl.jpg)](https://knowyourmeme.com/memes/disaster-girl)
+[![[disaster-girl.jpg]]](https://knowyourmeme.com/memes/disaster-girl)
 
 It is also posible to add images inline (example, here's the site favicon: [![[favicon.svg]]]{.w-6}) say in the middle of a paragraph.
 
--- a/docs/demo/embed.md
+++ b/docs/demo/embed.md
@@ -32,17 +32,13 @@ It is also posible to add images inline (example, here's the site favicon: [![[f
 
 ### Videos
 
-The following is the result of using `![[death-note.mp4]]`.
+The following is the result of using `![[death-note.mp4]]` (note that `![](death-note.mp4)` also works).
 
 ![[death-note.mp4]]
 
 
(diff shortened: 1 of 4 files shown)


srid:
Doing it this way means this syntax will also work for other formats, like video formats:

```markdown
![](foo.mp4)
```

mnaoumov:
@srid Maybe `Emanote.Pandoc.Link.InlineImage` name has to be changed then? Don't you think it is confusing now?

mnaoumov:
@srid maybe `InlineStaticFile` would be better ?

srid:
`InlineRef`, defined under `Emanote.Pandoc`, knows nothing about non-Pandoc specific types (like `StaticFile`). So I'd say it is fine as is. The type exists simply to distinguish between the two possible inline links allowed by Pandoc AST.

srid:
I actually should refactor the whole pandoc related rendering code one day. It has gotten way too complex. Meanwhile, I decoupled the Heist utilities to https://github.com/srid/heist-extra

mnaoumov:
@srid I understand that Pandoc mistakenly calls everything `Image` but I guess your should not keep using that leaking abstraction further in your library. What if you rename it to `InlineResource` and maybe write in the comment that Pandoc mistakenly calls `Text.Pandoc.Definition.Image` all resources including images, videos, pdfs etc? This will make your library's API clearer and leaked abstractions from Pandoc will be encapsulated within `Emanote.Pandoc.Link` module

srid:
> Pandoc mistakenly calls `Text.Pandoc.Definition.Image` all resources including images, videos, pdfs etc?

Pandoc is behaving correctly. Per commonmark spec, `` can only be used for images. 

https://commonmark.org/help/tutorial/08-images.html

It just so happens that Obsidian and the like decide to overload this syntax for embedding non-image resources.

If we want to improve the API properly, perhaps we should start from Pandoc itself (as its AST is not that flexible currently); eg.: https://github.com/jgm/pandoc-types/issues/98

mnaoumov:
@srid ok, I take my words back about misbehaving Pandoc. I'm so used to the embedded videos, audios, pdfs, so I forgot that pure Markdown still doesn't allow it

PR #356, opened 2022-10-17
Apply builtin Pandoc filters to backlink note excerpts

kukimik:
This addresses the issue I noticed in another PR https://github.com/EmaApps/emanote/pull/189#issuecomment-1279490368:

> I've just found out that the builtin filters are not applied to the note excerpts in the backlinks ("Links to this page") section. [...] this also means that tags are not converted to links and that the emoji don't get their `font-family: emoji styling`. Is this behavior intended? It looks like a bug.

Perhaps the approach in the following TODO would be better: https://github.com/EmaApps/emanote/blob/65dd75053674aaf4fc36c0ad689938c8d6158e2e/src/Emanote/Pandoc/BuiltinFilters.hs#L16

but it needs more time, and it's a TODO with a question mark, so there probably are some doubts here that I'm not aware of.

### DIFF
--- a/docs/guide/markdown.md
+++ b/docs/guide/markdown.md
@@ -115,7 +115,7 @@ Ordered lists,
 
 ## Hash Tags
 
-Add Twitter-like hashtags anywhere in Markdown file. They can also be added to the YAML frontmatter. Hash tags can also be "hierarchical", for instance: #emanote/syntax/demo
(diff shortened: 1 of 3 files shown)


srid:
Thanks, @kukimik. I'll review your two PRs as soon as I can. 

I'd also like to cut an official release (https://github.com/EmaApps/emanote/issues/153) soonish.

srid (approved):

Issue #360, opened 2022-10-23
Footnotes and embedding: duplicate HTML element ids 


kukimik:
A note that both

* contains a footnote, and 
* embeds another note, which also contains a footnote

will have two elements with the same id attribute in the resulting HTML, which is an error.

This happens because the footnotes numbering scheme does not depend on the note:

https://github.com/EmaApps/emanote/blob/944767a86152af3aab884e7c2ce5382c94162f33/default/templates/components/pandoc.tpl#L60

I think this can be resolved for example by including in some form (maybe a hash?) the note's route in the footnote's id. The downside is that the ids will be ugly.

Another idea is to construct the ids in a different way only when the note is embedded.

srid:
Related #110

PR #361, opened 2022-10-23
footnotes without javascript

kukimik:
This replaces the JS workaround of the `<base>` tag vs anchor links issue, and also prepares the ground for #137.

Be aware that this changes the behaviour of emanote when the user clicks a footnote link contained in an embedded note. However, footnotes in embedded notes are broken anyway #360 (and I think can be fixed in a reasonable way that is compatible with this PR), so maybe that's not a big issue.


srid (approved):
LGTM. Will merge soon.

PR #367, opened 2022-11-05
Make wikilink parser a separate library

srid:
https://github.com/srid/commonmark-wikilink


srid:
https://hackage.haskell.org/package/commonmark-wikilink-0.1.0.0

PR #369, opened 2022-11-06
Fix typo

tars0x9752:
Just found a typo.


srid:
This wasn't a typo, as `emanate` was a deliberate choice. I do see that this phrasing can be confusing, so I've clarified it in cc65b533e780f902ca223268c93f5423c489182a.

PR #376, opened 2022-11-14
Move tests to their own cabal component

srid:
Unbloat the emanote binary a bit.

- Introduce as well the an executable component
- Add a cabal flag to conditionally do things for ghcid mode (incremental compilation)


srid:
the ghcid flag trick is taken from https://github.com/srid/neuron/blob/master/neuron.cabal originally done in https://github.com/srid/neuron/pull/413 thanks to @Lysxia

PR #391, opened 2022-12-05
Add stork.js.map

kukimik:
The [source map](https://firefox-source-docs.mozilla.org/devtools-user/debugger/how_to/use_a_source_map/index.html) file `stork.map.js` referenced to in  https://github.com/EmaApps/emanote/blob/a688aaeef0819071c1466f38d03bfa454ec9003f/default/_emanote-static/stork/stork.js#L2 is missing. This causes a [source map error](https://firefox-source-docs.mozilla.org/devtools-user/debugger/source_map_errors/?utm_source=mozilla&utm_medium=firefox-console-errors&utm_campaign=default#source-map-missing-or-inaccessible) when viewing the site.

1. Go to https://emanote.srid.ca using Firefox (that's what I use; I'm sure you can observe it in other browsers too).
2. Open the developer tools..
3. Switch to the debugger tab. 
4. Observe the error report.

I fix this by adding the `stork.js.map` file downloaded from https://github.com/jameslittle230/stork/releases/tag/v1.5.0. The file [wasn't there](https://github.com/jameslittle230/stork/issues/322) when stork was included in emanote; that's probably the reason it is missing.


srid:
Ah, great. Looks like I should cut a 1.0.1.0 release with this fix (and potential others).

srid (changes requested):
The new file should also be added to the `.cabal` file:

https://github.com/EmaApps/emanote/blob/9deff5c4f890eb84494420b486e9498d1c001359/emanote.cabal#L27-L29

kukimik:
Thanks for pointing that out! I'll try not to forget this next time.

Do I also need to bump the version? I don't know what's the policy.

srid (approved):
Yea I see this in Brave for emanote.srid.ca

image

srid:
> Do I also need to bump the version? I don't know what's the policy.

There is no policy so far, and no obligation on PR author to do these. I usually update changelog and increment version once the PR is close to be ready to be merged, assuming it will be a squash merge. But if you want to update them yourself, that's fine too. I haven't really settled on versioning policy yet, suggestions welcome.

PR #396, opened 2022-12-08
Workaround for Prism.js and Tailwind CSS both using `table` class (#320)

### LINKED ISSUE #320: PrismJS-highlighted TOML table is wrong, due to Tailwind

kukimik:
This is a hack, but it looks like it works. If `table` turns out to be the only name clash, this workaround may be sufficient to close #320. But I won't mind if you prefer some more elegant solution and close this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   - Reduce Emanote's Nix runtime closure size
 - UI
   - Add source map for Stork [\#391](https://github.com/srid/emanote/pull/391)
+  - Workaround for Prism.js and Tailwind CSS both using `table` class [\#320](https://github.com/srid/emanote/pull/396)
 - Features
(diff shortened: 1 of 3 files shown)


srid (approved):
Since the scope of the hack is limited to `js.prism`, this is fine.

PR #398, opened 2022-12-10
Include YAML frontmatter in the Stork index

jfpedroza:
This PR sets the `frontmatter_handling` setting in Stork to `Ignore` so the frontmatter is included in the searchable text.

See #397

### DIFF
--- a/src/Emanote/Model/Stork/Index.hs
+++ b/src/Emanote/Model/Stork/Index.hs
@@ -13,7 +13,7 @@ module Emanote.Model.Stork.Index
 where
 
 import Control.Monad.Logger (MonadLoggerIO)
-import Data.Aeson (FromJSON (parseJSON))
+import Data.Aeson (FromJSON, genericParseJSON)
 import Data.Aeson qualified as Aeson
 import Data.Text qualified as T
 import Data.Time (NominalDiffTime, diffUTCTime, getCurrentTime)
@@ -92,7 +92,7 @@ data Handling
   = Handling_Ignore
   | Handling_Omit
   | Handling_Parse
-  deriving stock (Eq, Show)
+  deriving stock (Eq, Show, Generic)
 
 newtype Config = Config
   { configInput :: Input
@@ -140,9 +140,11 @@ configCodec =
   Config
     <$> Toml.table inputCodec "input" .= configInput
 
+handlingJSONOptions :: Aeson.Options
+handlingJSONOptions =
+  Aeson.defaultOptions
+    { Aeson.constructorTagModifier = toString . T.toLower . T.replace "Handling_" "" . toText
+    }
+
 instance FromJSON Handling where
-  parseJSON = Aeson.withText "FrontmatterHandling" $ \case
-    "ignore" -> pure Handling_Ignore
-    "omit" -> pure Handling_Omit
-    "parse" -> pure Handling_Parse
-    _ -> fail "Unsupported value for frontmatter-handling"
+  parseJSON = genericParseJSON handlingJSONOptions
--- a/src/Emanote/Model/Stork/Index.hs
+++ b/src/Emanote/Model/Stork/Index.hs
@@ -8,6 +8,7 @@ module Emanote.Model.Stork.Index
     readOrBuildStorkIndex,
     File (File),
     Input (Input),
+    Handling (Handling_Ignore, Handling_Omit, Handling_Parse),
   )
 where
 
@@ -19,7 +20,7 @@ import Numeric (showGFloat)
 import Relude
 import System.Process.ByteString (readProcessWithExitCode)
 import System.Which (staticWhich)
-import Toml (TomlCodec, encode, list, string, text, (.=))
+import Toml (Key, TomlCodec, diwrap, encode, list, string, table, text, textBy, (.=))
 
 -- | In-memory Stork index tracked in a @TVar@
 newtype IndexVar = IndexVar (TVar (Maybe LByteString))
@@ -59,7 +60,7 @@ storkBin = $(staticWhich "stork")
 
 runStork :: MonadIO m => Input -> m LByteString
 runStork input = do
-  let storkToml = handleTomlandBug $ Toml.encode inputCodec input
+  let storkToml = handleTomlandBug $ Toml.encode storkInputCodec $ StorkInput input
   (_, !index, _) <-
     liftIO $
       readProcessWithExitCode
@@ -79,8 +80,20 @@ runStork input = do
       -- title (but why would they?)
       T.replace "\\\\U" "\\U"
 
-newtype Input = Input
-  { inputFiles :: [File]
+data Input = Input
+  { inputFiles :: [File],
+    inputFrontmatterHandling :: Handling
+  }
+  deriving stock (Eq, Show)
+
+data Handling
+  = Handling_Ignore
+  | Handling_Omit
+  | Handling_Parse
+  deriving stock (Eq, Show)
+
+newtype StorkInput = StorkInput
+  { globalInput :: Input
   }
   deriving stock (Eq, Show)
 
@@ -98,7 +111,29 @@ fileCodec =
     <*> Toml.text "url" .= fileUrl
     <*> Toml.text "title" .= fileTitle
 
+showHandling :: Handling -> Text
+showHandling handling = case handling of
+  Handling_Ignore -> "Ignore"
+  Handling_Omit -> "Omit"
+  Handling_Parse -> "Parse"
+
+parseHandling :: Text -> Either Text Handling
+parseHandling handling = case handling of
+  "Ignore" -> Right Handling_Ignore
+  "Omit" -> Right Handling_Omit
+  "Parse" -> Right Handling_Parse
+  other -> Left $ "Unsupport value for frontmatter handling: " <> other
+
+handlingCodec :: Toml.Key -> TomlCodec Handling
+handlingCodec = textBy showHandling parseHandling
+
 inputCodec :: TomlCodec Input
 inputCodec =
   Input
-    <$> Toml.list fileCodec "input.files" .= inputFiles
+    <$> Toml.list fileCodec "files" .= inputFiles
+    <*> Toml.diwrap (handlingCodec "frontmatter_handling") .= inputFrontmatterHandling
+
+storkInputCodec :: TomlCodec StorkInput
+storkInputCodec =
+  StorkInput
+    <$> Toml.table inputCodec "input" .= globalInput
--- a/src/Emanote/Model/Stork/Index.hs
+++ b/src/Emanote/Model/Stork/Index.hs
@@ -110,51 +110,49 @@ data File = File
   }
   deriving stock (Eq, Show)
 
-fileCodec :: TomlCodec File
-fileCodec =
-  File
-    <$> Toml.string "path"
-      .= filePath
-    <*> Toml.text "url"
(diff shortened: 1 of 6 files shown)


src/Emanote/Model/Stork/Index.hs
```diff
@@ -79,11 +79,16 @@ runStork input = do
       -- title (but why would they?)
       T.replace "\\\\U" "\\U"
 
-newtype Input = Input
-  { inputFiles :: [File]
+data Input = Input
+  { inputFiles :: [File],
+    inputFrontmatterHandling :: Text
```
srid: Instead of `Text`, I'd create a new type here:

```haskell
data Handling 
  = Handling_Ignore
  | Handling_Omit
  | Handling_Parse
```

... to correspond to the 3 values in https://stork-search.net/docs/config-ref#frontmatter_handling

And then use it as:

```haskell
  inputFrontmatterHandling :: Handling 
```

src/Emanote/Model/Stork/Index.hs
```diff
@@ -59,7 +59,7 @@ storkBin = $(staticWhich "stork")
 
 runStork :: MonadIO m => Input -> m LByteString
 runStork input = do
-  let storkToml = handleTomlandBug $ Toml.encode inputCodec input
+  let storkToml = handleTomlandBug $ Toml.encode storkInputCodec $ StorkInput input
```
srid: Why this change?

I don't see any reason for adding `storkInputCode` and `StorkInput` either.

srid (changes requested):
Thanks for the PR. 

[Reading your comments here](https://github.com/EmaApps/emanote/discussions/397#discussioncomment-4357944), I'm convinced that we should disable this feature by default and add an option to `index.yaml` for manually enabling it. We can put the option under: https://github.com/EmaApps/emanote/blob/d2a64234fa327b782fcf7e49097226bea90248fe/default/index.yaml#L131-L133 

viz:

```yaml
emanote:
  stork:
    frontmatter-handling: ignore
```

src/Emanote/Model/Stork/Index.hs
srid: Reason: https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/

Specifically,

> - **Let your datatypes inform your code, don’t let your code control your datatypes.** Avoid the temptation to just stick a `Bool` in a record somewhere because it’s needed by the function you’re currently writing. Don’t be afraid to refactor code to use the right data representation—the type system will ensure you’ve covered all the places that need changing, and it will likely save you a headache later.

src/Emanote/Model/Stork/Index.hs
jfpedroza: I tried using only `inputCodec`, using `input.frontmatter_handling` but the generated file looked like

```toml
input.frontmatter_handling = "Ignore"

[[input.files]]
...
```

And Stork didn't like it.

Using another function with `Toml.table` was the only way I got it to work.

I might have missed something, though. I'm quite new to Haskell.

src/Emanote/Model/Stork/Index.hs
jfpedroza: Makes sense.

jfpedroza (commented):
> I'm convinced that we should disable this feature by default and add an option to index.yaml for manually enabling it.

I'll explore that part of the code and see how I do.

src/Emanote/Model/Stork/Index.hs
```diff
@@ -98,7 +110,29 @@ fileCodec =
...
+
+parseHandling :: Text -> Either Text Handling
+parseHandling handling = case handling of
+  "Ignore" -> Right Handling_Ignore
+  "Omit" -> Right Handling_Omit
+  "Parse" -> Right Handling_Parse
+  other -> Left $ "Unsupport value for frontmatter handling: " <> other
+
+handlingCodec :: Toml.Key -> TomlCodec Handling
+handlingCodec = textBy showHandling parseHandling
```
jfpedroza: It seems Tomland needs you to specify how to parse even if you are only generating. If we don't want to write code for that, we could use

```haskell
handlingCodec = textBy showHandling (\_ -> error "Not implemented")
```
or
```haskell
handlingCodec = textBy showHandling(\_ -> Left "Not implemented")
```

But I'm not sure if that's how you usually handle these cases.

jfpedroza:
It now reads the config from `index.yaml`, defaulting to "omit", the same default from Stork.

Tangential, but if I change the `index.yaml`, the index isn't rebuilt. I have to write a Haskell or Markdown file for it to be rebuilt. Is that a bug?

src/Emanote/Model/Stork/Index.hs
srid: That's fair, I'll take a look.

src/Emanote/Model/Stork/Index.hs
srid: Partial functions should be avoided as much as possible. More code is not necessarily bad. Here, Tomland is a *bidirectional* serializer, so it makes sense.

src/Emanote/Model/Stork/Index.hs
```diff
@@ -79,8 +82,20 @@ runStork input = do
...
+  deriving stock (Eq, Show)
+
+data Handling
+  = Handling_Ignore
+  | Handling_Omit
+  | Handling_Parse
+  deriving stock (Eq, Show)
+
+newtype StorkInput = StorkInput
+  { globalInput :: Input
```
srid: Maybe, `newtype Config` and `configInput` (to keep with the field naming convention).

src/Emanote/Model/Stork/Index.hs
```diff
@@ -98,7 +113,36 @@ fileCodec =
...
+
+parseHandling :: Text -> Either Text Handling
+parseHandling handling = case handling of
+  "Ignore" -> Right Handling_Ignore
+  "Omit" -> Right Handling_Omit
+  "Parse" -> Right Handling_Parse
+  other -> Left $ "Unsupported value for frontmatter handling: " <> other
+
+handlingCodec :: Toml.Key -> TomlCodec Handling
+handlingCodec = textBy showHandling parseHandling
```
srid: Is there no way to implement this codec in terms of `Show` instance?

src/Emanote/Model/Stork/Index.hs
```diff
@@ -98,7 +113,36 @@ fileCodec =
...
+storkInputCodec =
+  StorkInput
+    <$> Toml.table inputCodec "input" .= globalInput
+
+instance FromJSON Handling where
+  parseJSON = Aeson.withText "FrontmatterHandling" $ \case
+    "ignore" -> pure Handling_Ignore
+    "omit" -> pure Handling_Omit
+    "parse" -> pure Handling_Parse
+    _ -> fail "Unsupported value for frontmatter-handling"
```
srid: Is there no way to implement this instance in terms of `Show` instance?

srid (commented):
I left some comments indicating the changes the need to be done on top of this PR. I can do them myself, unless you want to give it a go.

> Tangential, but if I change the `index.yaml`, the index isn't rebuilt. I have to write a Haskell or Markdown file for it to be rebuilt. Is that a bug?

It normally should, IIUC. What browser you are on?

src/Emanote/Model/Stork/Index.hs
jfpedroza: This would be in terms of the `Read` instance, no?

src/Emanote/Model/Stork/Index.hs
jfpedroza: Tomland supports using `Show` and `Read`. They would need to be manually implemented instead of derived, so it uses `"Omit"` instead of `"Handling_Omit"`. Not much difference, but I guess it's the correct way.

src/Emanote/Model/Stork/Index.hs
jfpedroza: Stork needs the string capitalized. To reuse the instance, we would need to take the value capitalized in the YAML as well.

jfpedroza:
> It normally should, IIUC. What browser you are on?

Firefox.

src/Emanote/Model/Stork/Index.hs
srid: I think there should be a way to do this using generics. Aeson has generic deriving, even with options.

jfpedroza: I see. I haven't read about generics yet. I'll see how it goes.

src/Emanote/Model/Stork/Index.hs
srid: If we can implement Aeson instance generically, perhaps it would be simpler to just derive the Tomland codec via that Aeson instance. No need to modify the Show/Read instance. 

Just an idea; I haven't looked deep into how Tomland does things. You could also try asking in https://github.com/kowainik/tomland for idiomatic ideas ...

src/Emanote/Model/Stork/Index.hs
jfpedroza: I was able to derive the `FromJSON` instance generically, but Tomland doesn't support generic derive for sum types. I left the current implementation for now.

jfpedroza:
Changed `StorkInput` to `Config` and added some docs. I think it's ready.

srid:
bors try

srid:
bors try

srid:
@jfpedroza How can I manually test this PR?

src/Emanote/Model/Stork/Index.hs
```diff
@@ -91,18 +102,57 @@ data File = File
...
+  def = Handling_Omit
+
+instance FromJSON Handling where
+  parseJSON = genericParseJSON handlingJSONOptions
+    where
+      handlingJSONOptions :: Aeson.Options
+      handlingJSONOptions =
+        Aeson.defaultOptions
+          { Aeson.constructorTagModifier = toString . T.toLower . T.replace "Handling_" "" . toText
+          }
```
srid: This instance can be simplified using 

https://hackage.haskell.org/package/aeson-casing-0.2.0.0/docs/Data-Aeson-Casing.html

srid: @jfpedroza Wanna take a stab at this? 

Otherwise, I'll do the rest of the changes (including this one) tomorrow and merge.

srid:
bots try

srid:
bors try

src/Emanote/Model/Stork/Index.hs
jfpedroza: Yes, I'll give it a try.

jfpedroza: The package works on `fieldLabelModifier`, not `constructorTagModifier` so I can't use the public functions. And even with the internal functions, they don't work on the form `Handling_Ignore`, they expect camel case.

The only thing useful is `applyFirst` that allows you to do

```haskell
Aeson.constructorTagModifier = applyFirst toLower . drop 9
```

It's not worth to add the package for that function. I added it and changed `replace` with `drop` since it avoids the converting to text and back dance.

srid: Right. 

I actually should have suggested https://hackage.haskell.org/package/deriving-aeson because that's exactly what we need here.

cf. https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/deriving_via.html

srid:
bors try

PR #437, opened 2023-05-17
Upgrade haskell-flake (`packages-ng` branch)

srid:
Trying out https://github.com/srid/haskell-flake/pull/162 in emanote


flake.nix
```diff
@@ -37,50 +38,62 @@
...
+          in lib.mapAttrs
+            (_: _: {
+              settings = {
+                haddock = false; # Because, this is end-user software. No need for library docs.
+                libraryProfiling = false; # Avoid double-compilation.
+                # TODO: Must enable only if there is executable stanza.
+                justStaticExecutables = true; # Avoid needless runtime deps.
+              };
+            })
+            locals;
```
srid: @roberth So here's the current API for https://github.com/srid/haskell-flake/pull/162

It works pretty well. One problem is 'infinite recursion' if I address the first TODO by using `config.packages` (filtering over `localTo` to find local packages). 

One way to address this is to restructure `packages` like this:

```nix
{ 
  packages = { pkgs, self, super }: {
    sources = {
      emanote.root = ./.;
    };
    settings = {
      emanote.check = true;
    };
  };
}
```

Which is slightly complex, because now package-specific configuration is spread across two submodules - `sources` and `settings` - but OTOH I believe this will eliminate the infinite-recursion issue. 

I wonder if there are other ideas to consider.

roberth: Perhaps you could use `passthru` to store which flake a package came from, so that you direct access to this information when you need to get the locally defined packages from the set of by-source packages. Not sure if that helps, but perhaps doing it this way would allow you to think of these sets of packages in a way that you can untangle that recursion.
Just an idea, I'd have to dig through the code more to really understand the problem.

srid: I have no idea how `passthru` will help here. Simply referencing `config.packages` will trigger infinite recursion; I mean it makes sense that it does because we are building `config.packages` by reading `config.packages` (ouroboros).

Roughly, there are three layers of information here:

1. Package source info (path, executable stanzas, etc.)
2. Cabal override settings (doCheck, etc.)
3. Final nix package

The definition of (2) can depend on (1), specifically in these two cases:

- Set default overrides for a package if it is local to current flake (as this diff above shows)
- Enable `justStaticExecutables` on only local packages that have executables (as the 2nd TODO indicates).

This is reason for keeping (1) and (2) separate, because then (2) can be defined in terms of (1) without infinite recursion. But I'm probably missing a better design here ...

roberth: So for the module of type (2) to be composable anywhere it must not `imports` the module of type (1), but rather the information from (1) has to be translated to (2) in such a way that it can be used anywhere.
The flake identity in the passthru only comes in when going from (2) back to the once again local context of (3): filtering away the irrelevant packages.

So (1) could be a mere top level submodule about where the files are. There will be some translation from (1) to (2), where (2) is all about reusability, so deferredModule. Unlike (2), (1) is not exposed.
(3) is not reusable, but rather again a submodule, in which modules of class (2) can be imported.

(1)'s definition of packages is little more than pairs of names and paths.
(2)'s definition of packages is the most advanced: an overlay or something similarly expressive: something containing an overlay or something that can be translated to an overlay
(3)'s definition of packages is basically that of the Nix CLI or flake-parts `perSystem.packages`. All the composition has been done at the (2) level; it only needs to add the packages that are local to the flake outputs.

`passthru` would only serve to carry information from (1) to (3) without complicating (2).

roberth: > * Set default overrides for a package if it is local to current flake (as this diff above shows)
> * Enable `justStaticExecutables` on only local packages that have executables (as the 2nd TODO indicates).

These seem like responsibilities for (3) instead of (2), or you could put them in (2) anyway and make them conditional based on the flake identity in the `passthru`.

(3) would get to import modules from (2), and perhaps have one or two extra options for local concerns. It also has the effect of writing to the flake outputs. Ability to do `imports` can be represented by having the same [`class`](https://nixos.org/manual/nixpkgs/unstable/#module-system-lib-evalModules-param-class) (which is a new concept since like a week or two, not reflected in flake-parts yet).

srid: @roberth Say 'foo' project depends on 'bar'. If bar's (1) is not exposed, then how you can you import the local packages of 'bar' (defined only in (1)) in 'foo' ?.

I'd think that both (1) and (2) can be exported / imported. 

Also, (3) (`outputs.packages`) is currently an internal readOnly option.

srid: As of 2023d676d5c5cae8b7ebeadcd9500c0be5316abe, it uses the split `packages` and `settings` -- (1) and (2) -- of https://github.com/srid/haskell-flake/commit/8c018847b31be182cbdb673cb9904fb9f0524d2a

The project module `output` exports both (1) and (2), they are lazy attrs of deferred modules.

EDIT: The passthru became redundant this way.

srid: Now of course the new problem is that `apply` is not memoized, so it is invoked every time you reference `config.{packages,settings}`.

EDIT: I'm wrong; it is already memoized.

roberth: > Say 'foo' project depends on 'bar'. If bar's (1) is not exposed, then how you can you import the local packages of 'bar' (defined only in (1)) in 'foo' ?.

The information from (1) is automatically translated into the format of (2) and (3). It wouldn't make sense to have a single root directory when multiple flakes are loaded, eaching having their own root. That's why (1) is only loaded internally and translated to (2). (1) and its translation to (2) is basically just the directory traversal and the cabal parsing I think. So (1) only has very simple options. (2) is as expressive as an overlay, or perhaps more expressive. (3) is responsible for "calling" the overlay-like abstraction of (2), but it achieves that by being a submodule instead of a deferredModule. The options of (2) are present in (3).
A deferredModule always behaves like a function, even though its syntax may look like an attrset. That syntax is just sugar for the constant function case. A submodule on the other hand is evaluated (as in `evalModules`): it turns into an actual `config` attrset, which can be read by other options and/or by the caller of `evalModules` (when the caller reaches deep).

I have to read up a bit more before I can comment on the rest.

flake.nix
```diff
@@ -37,50 +38,63 @@
...
-                  ])
-                ];
+
+          packages = {
+            commonmark-extensions.root = "0.2.3.2";
+            # ema.root = lib.mkForce "0.8.2.0"; #  lib.mkForce (inputs.ema2 + /ema);
+          };
+          settings = {
+            emanote = { name, self, super, pkgs, ... }: {
+              check = false;
```
srid: Is it possible to move these args to the top-level? viz.:

```nix
settings = { name, self, super, pkgs, ... }: {
  emanote = { 
    check = false; ...
```

I don't believe so, because AFAIU a module cannot have dynamic list of top-level options.

PR #439, opened 2023-05-24
Fix H2 styling & Add documentation

bolt12:
See https://github.com/srid/emanote/discussions/438

I fix the styling on H2 heading component and attempt to add a section about how to override default styling.

### DIFF
--- a/docs/guide/markdown/custom-style.md
+++ b/docs/guide/markdown/custom-style.md
@@ -45,6 +45,35 @@ Using [fenced_divs](https://github.com/jgm/commonmark-hs/blob/master/commonmark-
 ![[file-links]]
 :::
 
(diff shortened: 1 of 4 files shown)


srid (approved):
Great, thanks.

bolt12:
My pleasure!

kukimik:
I'm a bit late, but this already was described, at least in part, here: http://emanote.srid.ca/guide/html-template.

I believe there should at least be a hyperlink between the two descriptions.

srid:
> I'm a bit late, but this already was described, at least in part, here: http://emanote.srid.ca/guide/html-template.
> 
> I believe there should at least be a hyperlink between the two descriptions.

I feel like we should just move it to the HTML template page.

PR #444, opened 2023-06-02
Added feature to embed code files

bolt12:
See https://github.com/srid/emanote/discussions/438#discussioncomment-6051427


srid:
Thanks! I'll take a look later (busy with other things), but just an initial question: can't the embedding itself happen client-side (by the highlighter library, for instance)?

This way they can do more cool stuff, like highlighting particular lines if asked for (cf. https://github.com/srid/neuron/issues/613).

bolt12:
As far as I understood for `highlight.js` to work the embedding (the file content) must already exist in the client side. `hljs.highlightAll()` will then look for all `pre code` nodes and highlight those.

If one would like to highlight particular lines if asked for one would have to parse those instructions and then fetch only the desired lines from the file contents, make them available in the DOM and then highlight them.

But I might be mistaken and there could be better ways to do this. I just followed what felt like the most natural way given how other embeddings work in Emanote

bolt12:
bump! Don't mean to stress you, but would love some feedback or maybe discuss this feature addition 😄

srid:
Sorry I didn't realize I had let 2 weeks slip by. Been a hectic recent few weeks.

My only concern is the use of IO during *rendering* stage. Emanote follows [Ema](https://ema.srid.ca/)'s design principle which is to pre-load all the data required _ahead_ in the `Model` type, and _then_ do the rendering from that. One way to approach this feature is to pre-`readFileText` all the "code" files linked from .md notes and store them in `Model`; then do a Map lookup in place of `unsafePerformIO` (that function, `embedStaticFileRoute`, already receives the `model` value as an argument).

srid:
Basically you want to begin by adding a new record field in this type to contain all these code snippets (that are referenced using embed wiki links in all .md notes):

https://github.com/srid/emanote/blob/5b24bd04f94e03afe66ee01da723e4a05d854953/emanote/src/Emanote/Model/Type.hs#L51-L69

srid:
Maybe `IxTask` is a good place to start, as to see how this could look (`IxCodeFiles`?).

bolt12:
Thank you @srid ! That makes sense, I'll look into it!

bolt12:
@srid Thanks for your pointers! Indeed this feels much cleaner. Let me know what you think of my approach.

srid:
Yea, that's certainly better. The information about memory caching of code file content is being spread across various modules like `Patch.hs`, so to avoid that kind of [complecting](https://www.infoq.com/presentations/Simple-Made-Easy/), we can create a type like this in `StaticFile.hs`:

```haskell
data StaticFileInfo where
  StaticFileCode :: StaticFileInfo Text  -- File content is embedded in the HTML itself
  StaticFileImage :: StaticFileInfo ()
  StaticFileAudio :: StaticFileInfo ()
  StaticFileVideo :: StaticFileInfo ()
  StaticFilePDF :: StaticFileInfo ()

data StaticFile = StaticFile
  { ...
  , staticFileInfo :: Maybe (DSum StaticFileInfo Identity)
  }

readStaticFileInfo :: FilePath -> IO (Maybe (DSum StaticFileInfo Identity))
readStaticFileInfo = 
  undefined
```

This way even the `extensionToTemplate` parsing of file exts is encapsulated in one place, `readStaticFileInfo`. 

Overall, we identify `StaticFileInfo` as [an area of volatility](https://www.informit.com/articles/article.aspx?p=2995357&seqNum=2) thus simplifying the code in the process.

bolt12:
That sounds great @srid thank you! And love the pointers/resources keep them coming, I wasn't aware of these 😄 .

I like the idea of using `DSum` to lift the file info to the type level and make the types more specific down the stream! However It seems `Eq`, `Ord`, `Show` and `ToJSON` deriving breaks when adding a `DSum` field to `StaticFile`. Any suggestions on how to work around that?

bolt12:
I went for a non dependent typed approach which I believe looks clean and simple enough too! Let me know what you think :)

srid:
This looks great! Could you also:

- Add an example to docs ([here](https://emanote.srid.ca/guide/markdown/embed))
- Run `treefmt` from `nix develop` shell so the code is autoformatted and hlint's are applied (if you don't have Nix installed, don't worry - I can do this once the PR is finalized with no pushes to be done). This will make the CI pass.

emanote/src/Emanote/Model/StaticFile.hs
```diff
@@ -37,4 +40,45 @@ staticFileSelfRefs =
     . R.unRoute
     . _staticFileRoute
 
+data StaticFileInfo where
+  StaticFileImage :: StaticFileInfo
+  StaticFileAudio :: StaticFileInfo
+  StaticFileVideo :: StaticFileInfo
+  StaticFilePDF   :: StaticFileInfo
+  StaticFileCode  :: Text -> StaticFileInfo
+  deriving stock (Eq, Show, Ord, Generic)
```
srid, line 56: Best to follow the naming convention of the project, so `StaticFileInfoImage` instead of `StaticFileImage`.

bolt12:
Done!

srid:
The margins are somewhat large, I'll try to reduce them:

image

bolt12:
Thank you, I am not very CSS savvy so I didn't bothered. If you can improve it, please do! Also as you can see Haskell files are sometimes misinterpreted, that's highlight-js fault too

srid:
Oh I just spotted something. @bolt12 The Haskell example is detected as `bash` (see screenshot). I think this is because your generated code block doesn't have a `haskell` class. Could you address this? Basically I suppose you want a `.hs -> haskell` mapping and put that as class name in code block. Use `bin/run` script to get live feedback on your changes.

Don't rebase; you can just add commits on top; I'll squash merge in the end anyway.

image

emanote/src/Emanote/Model/StaticFile.hs
```diff
@@ -37,4 +40,63 @@ staticFileSelfRefs =
...
+      , ".cpp"
+      , ".cs"
+      , ".rb"
+      , ".go"
+      , ".swift"
+      , ".kt"
+      , ".rs"
+      , ".ts"
+      , ".php"
+      ]
```
srid: Perhaps this should be a `Map FilePath Text`, viz.:

```haskell
codeExts = Map.fromList
  [ (".hs", "haskell") ]
```

bolt12:
@srid  yes I noticed that (see my comment). When I was tinkering with the code I tried exactly that, however those classes are added by highlight-js. It doesn't matter if I added them manually, at least that was what I discovered.

It seems Haskell files get misinterpreted as bash for some reason

srid:
> Also as you can see Haskell files are sometimes misinterpreted, that's highlight-js fault too

For highlight-js or any other syntax highligher to recognize it, you want to add the language name as code block class, like here:

image

https://raw.githubusercontent.com/srid/emanote/master/docs/tips/js/syntax-highlighting.md

srid:
> When I was tinkering with the code I tried exactly that

Hmm, I don't understand. I tried this:

```diff
--- a/emanote/default/templates/filters/embed-code.tpl
+++ b/emanote/default/templates/filters/embed-code.tpl
@@ -1,7 +1,7 @@
 <div class="py-0.5 mb-3 text-sm">
   <pre>
-    <code>
+    <code class="haskell">
 <ema:content />
     </code>
   </pre>
```

And it fixed the syntax highlighting issue on the Haskell code block for me. Basically, the fix here is to set this class dynamically, based on the file extension.

bolt12:
hmm okay, I'll give this a second try!

srid:
Yes, highlightjs adds its own classes, but it does based on the `class="haskell"` we add here (this is usually done in Markdown, but for this PR we have to do it in the `.tpl` file).

srid:
I think the class name should basically match the "alias" column here: https://github.com/highlightjs/highlight.js/blob/main/SUPPORTED_LANGUAGES.md

These values tend to be fairly standard/obvious, so they work across syntax highlighters.

bolt12:
Here's what I see:
image

As you can see the result is the same :/

bolt12:
On Chrome seems to look as it should

srid:
> On Chrome seems to look as it should

You mean on Chrome it is highlighted as Haskell, but on the other browser (which is?) that your screenshot shows above it is not being highlighted as Haskell (but rather as bash) despite setting the class name?

bolt12:
Yes exactly! The other browser is Firefox

srid:
Okay. I say, if it works in Chromium-based browsers, let's ship it. For Firefox, let's note this as a "known limitation" of highlightjs somewhere in the docs, ideally linking to the mozilla bug if one exists.

bolt12:
That sounds reasonable :) I'll push my changes

bolt12:
Done :)

emanote/src/Emanote/Model/StaticFile.hs
```diff
@@ -74,29 +84,31 @@ readStaticFileInfo fp readFilePath = do
           pure $ Just StaticFileInfoAudio
       | extension == "pdf" ->
           pure $ Just StaticFileInfoPDF
-      | extension `elem` codeExts ->
-          readFilePath fp <&> Just . StaticFileInfoCode
+      | extension `elem` Map.keys codeExts ->
+          readFilePath fp <&> Just . StaticFileInfoCode (codeExts Map.! extension)
```
srid: A `Map.lookup` (instead of list lookup via `elem` on `Map.keys`) here would be more efficient right?

bolt12: That's true, I forgot there's `Map.member`

bolt12: Fixed

srid (approved):
Functionality and code looks good. 👍🏿  I'll tweak a few things (CSS, template var names) and then merge latter.

bolt12:
Thank you very much for your help and guidance!

Issue #167, opened 2021-10-06
Disable hashtag in link text


srid:
> I notice my tags list is flooded with tags of the form #[\d+], coming from the titles of issue tracker links. Looking at the source I see the hashTagSpec follows CommonMark's allowing other inline elements in link titles, ie [*ital* **bold** #hash](foo). Still, in the case of hashtags my gut says they should be disallowed in the titles of other links (mostly because I want to be able to drop links from the browser without sanitizing the titles), also Obsidian doesn't parse them. WDYT?

_Originally posted by @edrex in https://github.com/srid/emanote/discussions/166_

edrex:
Reading WikiLink.hs, I see that wiki links are also parsed in markdown link text. BracketedSpec [disallows nesting](https://hackage.haskell.org/package/commonmark-0.2.1/docs/src/Commonmark.Inlines.html#line-829) links in the link text, but I'm not sufficiently familiar with Haskell or parsers to clearly see how to similarly disallow nesting WikiLinks and HashTags

srid:
FTR, using `\` will disable parsing of that inline tag, eg: `\#foo`. See the commit above.

Obviously, the ideal way is to have a way to control this globally so that these `\` are not required.

RoyiAvital:
Maybe add a logic which parse `#SomeText` in the following cases:

1. It is a beginning of a new row.
2. No space between `#` and the text.
3. No text after the adjacent text unless it is another `#` (Support list of tags).

If we want to follow [Obsidian Tags](https://help.obsidian.md/How+to/Working+with+tags) then no spaces in the tags as well and a tag can not be composed of digits only.

kukimik:
> Reading WikiLink.hs, I see that wiki links are also parsed in markdown link text.

Related: #349

Note that the embedding wikilinks (at least those that embed images) should be parsed in the link text.

----------

I tried implementing an workaround using a Pandoc filter that converts the tag `Span`s nested in link descriptions to their `Inlines` contents, but , but run into some obstacles, like handling of:

```
[see *issue #12345* on github](https://github.com/srid/emanote/12345)
```

The problem is that the second `*` is parsed as part of the hashtag.

Another nasty example:

```
[#abc{.someclass}](def)
```

On the other hand, as far as I understand, the discussion in https://github.com/jgm/commonmark-hs/issues/97 implies that this may be difficult to handle during parsing using `commonmark-hs`. Perhaps switching to Pandoc parser #264 is the way to go?

----------
[...]

PR #446, opened 2023-06-26
run `preparePandoc` in `parseNoteOrg` and `parseNoteMarkdown`

kukimik:
This implements the `TODO` from https://github.com/srid/emanote/blob/d165171ce46a967b0a72db138e06d16d86ef5dbb/emanote/src/Emanote/Pandoc/BuiltinFilters.hs#L16

I needed this for my failed approach at #167. I think this does not break anything and moreover it even fixes a bug:  the `font-family: emoji` style was not applied to the emoji in wikilink descriptions (this happened both in note body and in backlinks). See the screenshot of http://emanote.srid.ca/guide:

obraz

### DIFF
--- a/emanote/emanote.cabal
+++ b/emanote/emanote.cabal
@@ -1,6 +1,6 @@
 cabal-version:      2.4
 name:               emanote
-version:            1.0.3.8
+version:            1.0.3.9
 license:            AGPL-3.0-only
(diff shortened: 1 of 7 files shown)


srid (approved):
This is a good change. I don't remember anymore why I wrote code that way, but something broke when I did this back then.

PR #451, opened 2023-07-11
Add atom feed export

TristanCacqueray:
This change implements a feed view for atom reader.

- [x] Add feed content from note doc
- [x] Document the feature

### DIFF
--- a/emanote/src/Emanote/Model/Note.hs
+++ b/emanote/src/Emanote/Model/Note.hs
@@ -36,12 +36,21 @@ import Text.Pandoc.Definition (Pandoc (..))
 import Text.Pandoc.Readers.Org (readOrg)
 import Text.Pandoc.Walk qualified as W
 
+data Feed = Feed
+  { _feedEnable :: Bool
+  , _feedTitle :: Maybe Text
+  , _feedLimit :: Maybe Word
+  }
+  deriving stock (Eq, Ord, Show, Generic)
+  deriving anyclass (Aeson.ToJSON, Aeson.FromJSON)
+
 data Note = Note
   { _noteRoute :: R.LMLRoute
   , _noteDoc :: Pandoc
   , _noteMeta :: Aeson.Value
   , _noteTitle :: Tit.Title
   , _noteErrors :: [Text]
+  , _noteFeed :: Maybe Feed
   }
   deriving stock (Eq, Ord, Show, Generic)
   deriving anyclass (Aeson.ToJSON)
@@ -57,6 +66,8 @@ type NoteIxs =
      WL.WikiLink
    , -- HTML route for this note
      R 'R.Html
+   , -- XML route for this note
+     R 'R.Xml
    , -- Ancestor folder routes
      RAncestor
    , -- Parent folder
@@ -75,6 +86,7 @@ instance Indexable NoteIxs Note where
       (ixFun $ one . _noteRoute)
       (ixFun $ toList . noteSelfRefs)
       (ixFun $ one . noteHtmlRoute)
+      (ixFun $ maybeToList . noteXmlRoute)
       (ixFun noteAncestors)
       (ixFun $ maybeToList . noteParent)
       (ixFun noteTags)
@@ -115,6 +127,17 @@ lookupMeta :: Aeson.FromJSON a => NonEmpty Text -> Note -> Maybe a
 lookupMeta k =
   SData.lookupAeson Nothing k . _noteMeta
 
+noteHasFeed :: Note -> Bool
+noteHasFeed = maybe False _feedEnable . _noteFeed
+
+queryNoteFeed :: Aeson.Value -> Maybe Feed
+queryNoteFeed meta = do
+  feed <- SData.lookupAeson Nothing (one "feed") meta
+  let title = SData.lookupAeson Nothing (one "title") feed
+  let enable = SData.lookupAeson False (one "enable") feed
+  let feedLimit = SData.lookupAeson Nothing (one "limit") feed
+  pure $ Feed enable title feedLimit
+
 queryNoteTitle :: R.LMLRoute -> Pandoc -> Aeson.Value -> (Pandoc, Tit.Title)
 queryNoteTitle r doc meta =
   let yamlNoteTitle = fromString <$> SData.lookupAeson Nothing (one "title") meta
@@ -147,6 +170,12 @@ queryNoteTitle r doc meta =
     withoutH1 x =
       x
 
+-- | The xml route intended by user for this note.
+noteXmlRoute :: Note -> Maybe (R 'R.Xml)
+noteXmlRoute note
+  | noteHasFeed note = Just (coerce $ noteHtmlRoute note)
+  | otherwise = Nothing
+
 -- | The HTML route intended by user for this note.
 noteHtmlRoute :: Note -> R 'R.Html
 noteHtmlRoute note@Note {..} =
@@ -160,6 +189,10 @@ noteHtmlRoute note@Note {..} =
 lookupNotesByHtmlRoute :: R 'R.Html -> IxNote -> [Note]
 lookupNotesByHtmlRoute htmlRoute =
   Ix.toList . Ix.getEQ htmlRoute
+
+lookupNotesByXmlRoute :: R 'R.Xml -> IxNote -> [Note]
+lookupNotesByXmlRoute xmlRoute =
+  Ix.toList . Ix.getEQ xmlRoute
 
 lookupNotesByRoute :: HasCallStack => R.LMLRoute -> IxNote -> Maybe Note
 lookupNotesByRoute r ix = do
@@ -241,8 +274,9 @@ mkEmptyNoteWith someR (Pandoc mempty -> doc) =
 mkNoteWith :: R.LMLRoute -> Pandoc -> Aeson.Value -> [Text] -> Note
 mkNoteWith r doc' meta errs =
   let (doc'', tit) = queryNoteTitle r doc' meta
+      feed = queryNoteFeed meta
       doc = if null errs then doc'' else pandocPrepend (errorDiv errs) doc''
-   in Note r doc meta tit errs
+   in Note r doc meta tit errs feed
   where
     -- Prepend to block to the beginning of a Pandoc document (never before H1)
     pandocPrepend :: B.Block -> Pandoc -> Pandoc
--- a/emanote/src/Emanote/Model/Type.hs
+++ b/emanote/src/Emanote/Model/Type.hs
@@ -21,6 +21,7 @@ import Emanote.Model.Link.Rel qualified as Rel
 import Emanote.Model.Note (
   IxNote,
   Note,
+  noteHasFeed,
  )
 import Emanote.Model.Note qualified as N
 import Emanote.Model.SData (IxSData, SData, sdataRoute)
@@ -242,6 +243,18 @@ modelLookupNoteByHtmlRoute r =
   Rel.resolvedRelTargetFromCandidates
     . N.lookupNotesByHtmlRoute r
     . _modelNotes
+
+modelLookupFeedNoteByHtmlRoute :: R 'R.Xml -> ModelT f -> Maybe Note
+modelLookupFeedNoteByHtmlRoute r model = case resolvedTarget of
+  Rel.RRTFound note
+    | noteHasFeed note -> pure note
+    | otherwise -> Nothing
+  _ -> Nothing
+  where
+    resolvedTarget =
+      Rel.resolvedRelTargetFromCandidates $
+        N.lookupNotesByXmlRoute r $
+          _modelNotes model
 
 modelLookupTitle :: LMLRoute -> ModelT f -> Tit.Title
 modelLookupTitle r =
--- a/emanote/src/Emanote/Model/Type.hs
+++ b/emanote/src/Emanote/Model/Type.hs
@@ -234,9 +234,18 @@ modelDeleteData :: R.R 'R.Yaml -> ModelT f -> ModelT f
 modelDeleteData k =
   modelSData %~ Ix.deleteIx k
 
-modelLookupNoteByRoute :: LMLRoute -> ModelT f -> Maybe Note
-modelLookupNoteByRoute r (_modelNotes -> notes) =
-  N.lookupNotesByRoute r notes
+modelLookupNoteByRoute :: (R.LMLView, LMLRoute) -> ModelT f -> Maybe (R.LMLView, Note)
+modelLookupNoteByRoute (view, r) (_modelNotes -> notes) = do
+  note <- N.lookupNotesByRoute r notes
+  case (view, noteHasFeed note) of
+    (R.LMLView_Atom, True) -> pure (view, note)
+    (R.LMLView_Atom, False) -> Nothing
+    (_, _) -> pure (view, note)
+
+-- | Like `modelLookupNoteByRoute`, but for `R.Html` routes only.
+modelLookupNoteByRoute' :: LMLRoute -> ModelT f -> Maybe Note
+modelLookupNoteByRoute' r =
+  fmap snd . modelLookupNoteByRoute (R.LMLView_Html, r)
 
 modelLookupNoteByHtmlRoute :: R 'R.Html -> ModelT f -> Rel.ResolvedRelTarget Note
 modelLookupNoteByHtmlRoute r =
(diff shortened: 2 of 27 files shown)


TristanCacqueray:
Hello @srid, hope you don't mind me trying to implement #394 . I have a couple of questions:

- how to pass the configuration such as feedName, feedUrl, feedQuery, ... ?
- how to get the absolute url of a note?

srid:
Hey!

> how to pass the configuration such as feedName, feedUrl, feedQuery, ... ?

Normally, you can put them in either YAML configuration (eg: `index.yaml`) or Markdown frontmatter, depending on the semantics. Are you implementing one global feed? Or support for multiple feeds? 

> how to get the absolute url of a note?

https://github.com/srid/emanote/blob/d1dd7ea4ad89dc74f8d4b90650e07e788b489bb6/emanote/src/Emanote/Route/SiteRoute/Class.hs#L153

EDIT: That gives you relative to root URL, you can pass domain itself in config.

TristanCacqueray:
I'd be happy to implement what you see fit. I think a global feed with a single query to get the timeline would be good for most cases.

Thanks, I'll look into adding a global feedQuery to the YAML config.

srid:
Multi-feed may be simpler to implement and is also future-proof (what if someone wants to start a second blog?).

Consider an example `blog.md` that has a query to show recent blog listing. Can we automatically produce a corresponding `blog.xml` Atom feed based off that query, if the user has asked for it in the frontmatter? viz:

~~~
$ cat blog.md
---
slug: blog
generate_feed: true
---

## Recent posts

```query
tag:blog
```
~~~

This would be similar https://neuron.zettel.page/feed

TristanCacqueray:
Alright, that sounds good. It's just that I don't know yet where to hook the `blog.xml` file when there is a `generate_feed: true` in the note :-)

srid:
You can perhaps start with adding a new field of type `Maybe Feed`:

https://github.com/srid/emanote/blob/d1dd7ea4ad89dc74f8d4b90650e07e788b489bb6/emanote/src/Emanote/Model/Note.hs#L39-L45

Then perhaps add a `ResourceRoute_Feed LMLRoute` constructor here:

https://github.com/srid/emanote/blob/d1dd7ea4ad89dc74f8d4b90650e07e788b489bb6/emanote/src/Emanote/Route/SiteRoute/Type.hs#L36-L38

TristanCacqueray:
Ok, thanks, I think I can do it that way.

Then I have another question about the html template, how to add a
```
<link href="${feed:feedurl}" type="application/atom+xml" rel="alternate" title="${feed:title}"/>
```
conditionally? It seems like I should add the values in `Emanote.View.Common.commonSplices`, but I'm not sure how to do the `if feed is defined; then add the <link>` in the `base.tpl`. Is this the intended way, or should it be a different template? I could use some help here too.

srid:
One approach is to hook into `page.headHtml`, but there may be better approaches (checkout how sidebar is rendered).

https://github.com/srid/emanote/blob/d1dd7ea4ad89dc74f8d4b90650e07e788b489bb6/emanote/default/templates/base.tpl#L22

I suppose we can tackle this latter seeing how the implementation shapes up.

TristanCacqueray:
I may have spoken too soon :-) I'm shooting in the dark here as I'm not familiar with GADTs or type families... If you could check the shape of 912127a, it would be good to know if I'm on the right path. It seems like I needed to add a new `'Xml` FileType to please ghc.

So If that looks good to you, then it seems like I can almost pass the note metadata to the renderFeed!

srid:
I'll take a look. Meanwhile, you'll find [this post](https://lexi-lambda.github.io/blog/2021/03/25/an-introduction-to-typeclass-metaprogramming/) quite useful to understand the foundations behind those concepts (GADTs, type families).

Also, once you get something working, we can then refactor the code latter.

TristanCacqueray:
@srid Thanks!

39ba3dc appears to work with the run command, e.g. adding to my blog.md:

```
---
slug: blog
generate_feed: true
feed_url: https://tristancacqueray.github.io
feed_name: Tristan's Zettelkasten
---

# Blog

My blog articles:

```query {.timeline}
tag:blog
```

I get the feed when running `curl http://127.0.0.1:8080/blog.xml`.
However the xml file is not available when running the `gen` command.

srid:
> However the xml file is not available when running the `gen` command.

You want to add your feed URLs to the return value of this function: https://github.com/srid/emanote/blob/d1dd7ea4ad89dc74f8d4b90650e07e788b489bb6/emanote/src/Emanote/Route/SiteRoute/Class.hs#L38-L39

(It is basically Ema's `routeUniverse`: https://ema.srid.ca/guide/route/class)

TristanCacqueray:
@srid Thanks! I updated my site with this change (see: https://github.com/TristanCacqueray/TristanCacqueray.github.io/commit/d4b0067968e0a3d97a03da2de35341c81a59ae29), and here is the new feed https://tristancacqueray.github.io/blog.xml :)

emanote/src/Emanote/View/Feed.hs
```diff
@@ -0,0 +1,51 @@
...
+import Emanote.Route.R (indexRoute)
+import Emanote.Route.SiteRoute
+import Optics.Operators ((^?))
+import Optics.Optic ((%))
+import Relude
+import Text.Atom.Feed qualified as Atom
+import Text.Atom.Feed.Export qualified as Export (textFeed)
+
+getNotes :: Text -> Model -> [Note]
+getNotes tagName model = case parseQuery ("tag:" <> tagName) of
```
srid: Instead of hardcoding the query, we should make that configurable in the frontmatter or -- better -- automatically use the first (and only) query in the body of the Markdown note.

emanote/src/Emanote/Model/Type.hs
```diff
@@ -243,6 +245,23 @@ modelLookupNoteByHtmlRoute r =
     . N.lookupNotesByHtmlRoute r
     . _modelNotes
 
+isNoteFeed :: Note -> Bool
+isNoteFeed note = case _noteMeta note ^? key "generate_feed" % _Bool of
+  Just True -> True
+  _ -> False
```
srid: Can you make this a bit static, by storing the flag in the `Note` record itself?

emanote/src/Emanote/Model/Note.hs
```diff
@@ -147,6 +150,16 @@ queryNoteTitle r doc meta =
     withoutH1 x =
       x
 
+-- | The xml route intended by user for this note.
+noteXmlRoute :: Note -> R 'R.Xml
```
srid: Shouldn't this return a `Maybe`, because not all notes will have a feed?

emanote/src/Emanote/Model/Note.hs
```diff
@@ -147,6 +150,16 @@ queryNoteTitle r doc meta =
...
+noteXmlRoute note@Note {..} =
+  -- Favour slug if one exists, otherwise use the full path.
+  case noteSlug note of
+    Nothing ->
+      R.withLmlRoute coerce _noteRoute
+    Just slugs ->
+      R.mkRouteFromSlugs slugs
```
srid: I don't think you need this logic; just call `noteHtmlRoute` and coerce its result.

emanote/src/Emanote/Route/Ext.hs
```diff
@@ -15,6 +15,7 @@ data SourceExt = SourceExt
 data FileType a where
   LMLType :: LML -> FileType SourceExt
   Yaml :: FileType SourceExt
+  Xml :: FileType SourceExt
```
srid: This should be `Xml :: FileType ()` since `.xml` is not a source file (it is rather an output file).

emanote/src/Emanote/Route/R.hs
```diff
@@ -88,6 +88,14 @@ encodeRoute (R slugs) =
   let parts = Slug.unSlug <$> slugs
    in withExt @a @ft $ toString $ T.intercalate "/" (toList parts)
 
+decodeXmlRoute :: FilePath -> Maybe (R 'Xml)
```
srid, line 91: Maybe generalize `decodeHtmlRoute` and use that?

emanote/src/Emanote/View/Feed.hs
```diff
@@ -0,0 +1,51 @@
...
+        noteTitle = (Atom.TextString (toPlain $ _noteTitle note))
+
+renderFeed :: UTCTime -> Model -> Note -> LByteString
+renderFeed now model baseNote = encodeUtf8 $ fromMaybe "<error>bad feed</error>" mFeedTxt
+  where
+    mFeedTxt = do
+      feedSlug <- meta ^? key "slug" % _String
+      feedUrl <- meta ^? key "feed_url" % _String
+      feedName <- meta ^? key "feed_name" % _String
+      let noteUrl note =
```
srid: Instead of,

```yaml
generate_feed: true
feed_url: https://tristancacqueray.github.io
feed_name: Tristan's Zettelkasten
```

let's go with:

```yaml
feed:
  enable: true
  url: https://tristancacqueray.github.io
  title: Tristan's Zettelkasten
```

Actually, what is `url` doing here? Shouldn't it be in `index.yaml`? If you have multiple feeds, they all have a common base URL.

srid (commented):
Not a *thorough* review, but I thought I'd give a quick feedback.

emanote/src/Emanote/Route/R.hs
TristanCacqueray: I'm not exactly sure how to do that.

emanote/src/Emanote/View/Feed.hs
TristanCacqueray: Perhaps the title could default to the _noteTitle too?

TristanCacqueray:
Thanks for the review! I'll address the remaining comments later.

emanote/src/Emanote/View/Feed.hs
TristanCacqueray: It doesn't seem like `index.yaml` already has a `url`, so I think we should fallback to the index only if the feed doesn't have one already. Could you please show how to get the `_noteMeta` of the `index.yaml` ?

srid: index.yaml can contain the defaults, which then gets merged with `.md`'s frontmatter YAML

This function automatically does the merge (using `SData.mergeAesons`):

https://github.com/srid/emanote/blob/84aaaac7384571576506377ccdd5182fc5662ac7/emanote/src/Emanote/Model/Meta.hs#L18-L21

https://github.com/search?q=repo%3Asrid%2Femanote%20lookupRouteMeta&type=code

emanote/src/Emanote/Route/R.hs
srid: `coerce` should work, eg.: https://github.com/srid/emanote/blob/84aaaac7384571576506377ccdd5182fc5662ac7/emanote/src/Emanote/Model/Type.hs#L134

srid: Okay, let's address this at the end.

emanote/src/Emanote/View/Feed.hs
TristanCacqueray: Alright thanks! So last commit pulls the url from either:

```
feed:
  url: X
```

Or from the index.yaml:

```
site:
  url: X
```

TristanCacqueray:
Hey @srid, thank you for the help, I think this PR is now ready for review. It doesn't seems like there is a test infrastructure I could use to validate the change, but beside that, the feature should be complete. Let me know if there is anything else I could do, or feel free to adapt the implementation to your taste.

TristanCacqueray:
> You'll find [this post](https://lexi-lambda.github.io/blog/2021/03/25/an-introduction-to-typeclass-metaprogramming/) quite useful to understand the foundations behind those concepts (GADTs, type families).

Thank you for the link. I think I read it before, but I haven't found a use for this yet. Though it's good to see this in action in emanote :-)

srid:
I made a typo in the YAML frontmatter ("enable" instead of "enabled") which caused Emanote to error out. A bug somewhere? I haven't looked in detail:

```diff
diff --git a/docs/guide/query.md b/docs/guide/query.md
index 2af625a8..188d946c 100644
--- a/docs/guide/query.md
+++ b/docs/guide/query.md
@@ -1,5 +1,7 @@
 ---
 order: -95
+feed:
+  enable: true
 ---
 
 # Obsidian-style queries
```

➡️

```
emanote: attempt to encode missing note: LMLRoute_Md R[/guide/query.md]
CallStack (from HasCallStack):
  error, called at src/Relude/Debug.hs:289:11 in relude-1.1.0.0-KCT8nCRmwYB8RNjx4L9ojh:Relude.Debug
  error, called at src/Emanote/Route/SiteRoute/Class.hs:115:10 in main:Emanote.Route.SiteRoute.Class
  encodeResourceRoute, called at src/Emanote/Route/SiteRoute/Class.hs:90:9 in main:Emanote.Route.SiteRoute.Class
  emanoteRouteEncoder, called at src/Emanote.hs:48:27 in main:Emanote
```

srid:
http://127.0.0.1:9010/guide/query.xml => `<error>index.yaml or note doesn't have url</error>`

This should cause static-site generation to fail, but currently it does not (because `nix build .#docs` succeeds). The static-site sanity checks are done around here: https://github.com/srid/emanote/blob/34412693d180b9da51109e204d19800f2655843c/emanote/src/Emanote.hs#L88C1-L89

TristanCacqueray:
Oops, the error was happening because the discovery link assumed the feed was enabled (it was also triggered by setting `enabled: false`). This should be fixed by e610b8a6.

Then 79bc64ef should make the generation fail on feed error.

srid:
There is still an issue here.

@TristanCacqueray Are you using Nix to develop this PR? [The CI failure](https://jenkins.srid.ca/job/Emanote/job/PR-451/18/console) is coming from `nix build .#docs` (I can also see the same error in local dev live server, run via `bin/run`).

image

TristanCacqueray:
@srid arg, I messed up the hlint suggestion, 306ffca should pass the ci.

srid:
Great. I'd suggest the following API:

```yaml
feed:
  enable: true  # Note: "enable", not "enabled"
template:
  siteUrl: https://emanote.srid.ca  # ie, alongside template.baseUrl
````

**Comment by srid** [comment]
Great. I'd suggest the following API:

```yaml
feed:
  enable: true  # Note: "enable", not "enabled"
  siteUrl: https://emanote.srid.ca 
````

**Comment by srid** [comment]
To clarify, you don't need `site.url`. If you use `getEffectiveRouteMeta`, looking up `feed.siteUrl` will automatically search up to the top-level `index.yaml` to find it.

**Comment by TristanCacqueray** [comment]
@srid alright, good suggestions. Last commit should implement the change.

**Comment by srid** [comment]
I found a couple of issues

1. Adding `feed.title` seems to have no effect, as the feed is still using the note title. 1d1d31f
2. It seems you can't link to the .xml in Markdown, emanote thinks it is broken link. Looks like the route decoder is not properly implemented? 2157e26

**Comment by TristanCacqueray** [comment]
1. should be fixed in b4c7ffa
2. It seems like the culprit comes from `mkModelRouteFromFilePath`. Do I need to add a ModelRoute_Feed constructor to ModelRoute ?

**Comment by srid** [comment]
> 2\. It seems like the culprit comes from `mkModelRouteFromFilePath`. Do I need to add a ModelRoute_Feed constructor to ModelRoute ?

Yea.

But this is making me think that if we probably should refactor the output route type as follows:

```haskell
data ResourceRoute
  = ResourceRoute_StaticFile StaticFileRoute FilePath
  | ResourceRoute_LML LMLRoute LMLView

data LMLView = LMLView_Html | LMLView_Atom
```

Then you can just reuse `LMLView` in the `ModelRoute`.

TristanCacqueray:
What about LMLRoute, do you think we need a `LMLRoute_Xml (R 'Xml)` constructor?

TristanCacqueray:
Note sure how to get the link to point at the actual xml resource, but e7484b9 fixes the link render error, but it now points at the html file. It seems like we need a new LMLRoute, but could you confirm?

TristanCacqueray:
@srid thinking about it, changing the ModelRoute to accommodate feed links doesn't fit very well. In the last commit 2815a24 I propose a different strategy, would that work for you?

TristanCacqueray:
@srid I've squashed the commits and removed the ModelRoute change (e.g. internal link to feed xml are not working). Would it be possible to do that in a followup, since the feature is working otherwise.

srid:
@TristanCacqueray It is best to fix this glitch before merging to `master`. We don't want surprises; in fact when I was testing your branch locally right now, I was surprised that the `query.xml` took me to some other domain.

 I've been meaning to do this myself, so I just took a brief look at it. I think you can solve this problem by replacing `(Either MN.Note SF.StaticFile)` in Resolve.hs with `(Either (MN.Note, LMLView) SF.StaticFile)` (or equivalent). And maybe `ModelRoute_LML LMLView LMLRoute` constructor as well in `ModelRoute`. Haven't had the time to take a deeper look ...

srid:
> And maybe `ModelRoute_LML LMLView LMLRoute` constructor as well in `ModelRoute`.

Just implemented this.

Aside: I also realized that it has become a pain to work with Emanote code. I find that it has gotten way too complex! Time for a rewrite? 🤔

TristanCacqueray:
@srid thank you for your help and for merging this feature! For what it's worth, I don't think Emanote is too complex, perhaps it deserves a rewrite, but can it be less complicated? My next wish is a better timeline as requested in [#422](https://github.com/srid/emanote/discussions/422), for example to show a reading time estimation based on the word count.

PR #482, opened 2023-12-19
Add flake template (emanote-template)

shivaraj-bh:
Documentation for https://github.com/srid/emanote-template/pull/68


shivaraj-bh:
I am thinking if this is the right place to document about templates. Also [emanote-template](https://github.com/srid/emanote-template) has a step in README to "use the template" via GUI on github, maybe this section can go into the README as an alternative way to create the template?

srid:
It can remain here, as a separate page, then that README can link to it. I'll do this.

srid:
@shivaraj-bh I'm gonna simplify things which also has the added advantage of keeping emanote-template's flake smaller.

PR #483, opened 2023-12-21
Folgezettel-aware sidebar + new configurable 'default' layout

### LINKED ISSUE #473: Folgezettel sidebar
Allow the sidebar to be defined based on the folgezettal heterarchy rather than folder hierarchy.

Prior discussion:
- https://github.com/srid/emanote/discussions/64
- https://github.com/srid/emanote/discussions/50

Current discussion:
- https://nixos.zulipchat.com/#narrow/stream/413951-general/topic/obsidian.20graph.20in.20emanote

srid:
Resolves #473 by making the sidebar tree be built based on not only the folder structure but also folgezettel relationships. 

Also replaces both 'book' and 'note' layout with a single 'default' layout in which all of sidebar, uptree and breadcrumbs can be toggled on or of.

Still exploring this feature ... 

- [x] Create a working version for nixos.asia
- ~~[ ] A notion of 'primary' parent?~~
  - [x] Breadcrumbs still use folder structure (see `routeBreadcrumbs`); what to use? (Answer: fine as it is for now)
  - ~~[ ] Do we trim excessive parents? https://mathworld.wolfram.com/MaximumSpanningTree.html~~
- [x] Change semantics of folder-folgezettel (reverse). Done ... does it make sense?
- [x] "Index" (-/all) should _always_ include all notes, even if they cannot be reached from top (due to folder-folgezettel being cut off at some point or due to disconnected graphs). Perhaps this can be implemented by having index folder folgezettel automatically include orphans somehow (if so, under what user config?) https://en.wikipedia.org/wiki/Spanning_tree
    - [x] Fix excessive use of bold in Index
    - [x] Avoid having to specify folder-folgezettel to false in index note (for that private notebook deployed to CF pages)
- [x] use a different svg icon than 'folder'?
- [x] Replace 'book' and 'note' layout with a single layout
- [x] Performance (memoize tree)
- [x] Refactor


srid:
First pass, showing multiple parents. Obviously the sidebar needs to be trimmed:

image

srid:
Using this in https://nixos.asia/en now as of https://github.com/nixos-asia/website/commit/90db854c29703e99299a3187241a18cb4670bed0. `en.md` specifies the top-level folgezettel relationship explicitly, so now our sidebar looks:

image

srid:
It is a bit jarring to see `flake-parts` appear (unnecessarily) in two sub-trees here: https://nixos.asia/en/treefmt

image

Issue #494, opened 2024-01-12
Combining multiple notebooks


srid:
Being able to combine multiple (otherwise disparate) notebooks is useful, such as to render documentation coming from different projects (à la https://community.flake.parts/modules). What does it take to do this in Emanote?

## Sub-tasks

- [x] Ambiguous notes https://github.com/srid/emanote/pull/498
  - For eg., multiple projects may have `guide/index.md` (or `guide.md`). If one project links to it as `[[guide]]`, how do we know to resolve the ambiguity here in the combined notebook?
    - One approach: automatically choose the note with the most recent common ancestor
- [x] Support `folder1/index.md` (in addition to `folder1.md`) for folder note. #512
  - @vcavallo, @kouloumos (https://github.com/srid/emanote/discussions/107) and @zimbatm (https://github.com/srid/emanote/discussions/203) have asked for this in the past
- [x] unionmount should support "mount at" folder feature.  https://github.com/srid/emanote/pull/523
  - This enables us to avoid the [manual cp to subdir](https://github.com/flake-parts/community.flake.parts/blob/87df2e429591d561226a1f7c1370c089d33a0e92/flake.nix#L50) hack
  - And it can enable including sub-notebooks in live server CLI invocation (which can't do the `cp` hack)
  - Furthermore, it will address the edit URL problem automatically. https://github.com/flake-parts/community.flake.parts/issues/20

## Stretch

- [ ] Slug resolution
  - This one's tricky; we could probably get by with just not supporting slug transformation in version 1.
- [ ] Hierarchical site title (https://github.com/flake-parts/community.flake.parts/issues/4)

srid:
Ambiguous note resolution works now. 

Now on to index routes support ...

srid:
`index.md` for folder notes are now supported: https://emanote.srid.ca/folder-note#index

One more thing to do: allowing layers to be mounted at a subfolder.

PR #498, opened 2024-01-13
Resolve ambiguities based on closer common ancestor

srid:
- [x] Prototype
- [ ] Do the reverse: 
  - [x] "backlinks" should not show all the places linking to `[[foo]]` (if that foo was an ambiguity resolved)
  - [ ] what about uptree? folgezettel parents, etc.
- [x] refactor


edrex:
I was just looking into how it might be possible to add resolution (nearest ancestor), since I have same-name notes in different directories all over the place, making the current behavior a blocker for using emanote for my big repo. So great to see a patch in progress Srid!

I would prefer that the behavior be to omit descendant search entirely, and search only ancestors (that's what Ikiwiki does), but I don't need that to get unblocked.

I just did a test build and it's still showing ambiguous link errors for me:

Screenshot-2024-01-18-12:56

(out of scope here, the rendering for these inline errors has always seemed broken for me. Missing style or something?)

LMK if you want review when you're ready.

edrex:
Previous discussion, starting at https://github.com/srid/emanote/discussions/115#discussioncomment-1149833

srid:
> I would prefer that the behavior be to omit descendant search entirely, and search only ancestors (that's what Ikiwiki does), but I don't need that to get unblocked.

Without descendant search, it would not be possible to "combine" multiple sites though (see #494), and thus it would break https://community.flake.parts/ (the various docs come from different repos; and `guide.md` for example exists in both haskell-flake and nixos-flake repos).

edrex:
I was too ambigious with "... omit descendant search entirely". I meant a ref `Foo` shouldn't resolve to `/A/Foo.md`, but (`A/Foo`) still would. So a ref R from path A would resolve to B if and only if `B = Apre + Ref` for some Apre in Prefixes(A).

This would require all descending path elements to be explicit in refs. This would allow matching on subdirs for flexible reference remapping (eg, a local site registry mapping ref prefixes to URL prefixes for external sites).

I'm only guessing at the use-case you're working on. Am I missing part of the requirement?

srid:
@edrex What is the value of `Prefixes(A)` if A = `/Foo/Bar/Qux.md`? 

If `Prefixes` == `["/", "/Foo", "/Foo/Bar"]`, then by the rule you stated above, Qux.md can link to **any** note in the notebook. 

Could you provide a more rigorous logic so I can understand what you are proposing?

srid:
I'm gonna merge this now; but will always be happy to revisit the ambiguity resolution algorithm based on convincing rationale.

PR #503, opened 2024-01-22
Update tailwind for liveserver

TristanCacqueray:
This change makes the liveserver work with the latest tailwind version, similarly to what the site export produces.


srid (approved):
Interesting, I didn't know this (JS-based) existed. Did they introduce it recently?

TristanCacqueray:
I don't know when that happened, that's what I used in [xstatic-tailwind](https://github.com/TristanCacqueray/haskell-xstatic/tree/main/xstatic-tailwind) 2 years ago.

PR #512, opened 2024-02-09
Folder notes can now live in `$folder/index.md`

srid:
For #494 

- [x] `foo/index.md` should be equivalent to `foo.md`
- [x] `foo/index.yaml` should be equivalent to `foo.yaml`
- [x] How to deal with broken edit button link for `index.md` notes?
- [x] Refactor


srid:
> * How to deal with broken edit button link for `index.md` notes?

Because, the source-path is now no longer accurate:

image

We can't use `encodeRoute` anymore - as the information about the `index.md` slug is lost.

PR #528, opened 2024-03-15
templates: prevent the external link icon from wrapping

TristanCacqueray:
See: https://stackoverflow.com/questions/16100956/prevent-after-element-from-wrapping-to-next-line


TristanCacqueray:
Before: 
image

After: 
image

PR #529, opened 2024-03-15
Remove footnotes from `og:description`

TristanCacqueray:
This change remove any footnotes from the og:description header.

### DIFF
--- a/emanote/src/Emanote/Model/Note.hs
+++ b/emanote/src/Emanote/Model/Note.hs
@@ -387,7 +387,7 @@ applyNoteMetaFilters doc =
            )
     addDescriptionFromBody =
       overrideAesonText ("page" :| ["description"]) $ \case
-        B.Para is -> [WL.plainify is]
+        B.Para is -> [WL.plainify (mapMaybe removeInlineNotes is)]
         _ -> mempty
     -- FIXME this doesn't take splice rendering into account. Specifically,
     -- `![[foo.jpeg]]` is not handled at all.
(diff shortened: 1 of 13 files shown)


TristanCacqueray:
I'm sure there is a better way to filter the pandoc ast, but that fixes my issue.

emanote/src/Emanote/Model/Note.hs
```diff
@@ -406,4 +406,10 @@ applyNoteMetaFilters doc =
               pure $ SData.oneAesonText (toList key) val
           )
 
+-- TODO: apply this recursively
```
srid: You can address this TODO using [`Text.Pandoc.Walk.walk`](https://hackage.haskell.org/package/pandoc-types-1.23.1/docs/Text-Pandoc-Walk.html#v:walk).

srid (commented):
@TristanCacqueray I think it is better to fix it at the source itself, so whoever calls this function will benefit:

https://github.com/srid/heist-extra/blob/6b609c347754b15d8796c3d70d0ef6fb0ada8ca5/src/Heist/Extra/Splices/Pandoc/Render.hs#L280-L294

What do you think?

TristanCacqueray:
Sure, thanks, that's much better!

srid:
@TristanCacqueray Have you tested it to work? Because it is not working for me:

image

TristanCacqueray:
I remember testing the change to heist-extra, but looking at the code it does not seem to make sense (the catchall is already removing unmatched inlines). First of all, the function comes from `commonmark-wikilink`, and using `query` does not seem to be enough as it descends the Note node even when it is replaced with mempty. So I think we still need to somehow filter that node from the tree.

TristanCacqueray:
@srid sorry about that, now using `commonmark-wikilink.url = "github:TristanCacqueray/commonmark-wikilink/plainify-without-notes";` seems to work for the adding-images description.

srid:
Something's still off, because the "Paste Image" part repeats at the end.

image

srid:
@TristanCacqueray It would be good to merge this, after that issue above (https://github.com/srid/emanote/pull/529#issuecomment-2018733384) is resolved.

TristanCacqueray:
@srid I agree, unfortunately I don't have much time to work on this at the moment.

PR #533, opened 2024-03-20
Fix flake.parts docs

roberth:
It was trying to access some unset values.

I haven't documented the "default" for the `output` values, because they're not really defaults. You could describe them in the description if you feel it's useful info.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,6 +33,9 @@ jobs:
       - name: Check closure size
         run: |
           nix run .#check-closure-size
+      - name: Docs for flake.parts
+        run: |
+          nix build github:hercules-ci/flake.parts-website#generated-docs-emanote --override-input emanote ../emanote/ --show-trace
 
   # TODO: Use self-hosted runner
   website:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -33,6 +33,9 @@ jobs:
       - name: Check closure size
         run: |
           nix run .#check-closure-size
+      - name: Docs for flake.parts
+        run: |
+          nix build github:hercules-ci/flake.parts-website#generated-docs-emanote --override-input emanote ../emanote/ --show-trace
```
srid: Are there any docs on this? 

haskell-flake does something else entirely: https://github.com/srid/haskell-flake/blob/ef955d7d239d7f82f343b569a4cf2c7c1a4df1f4/doc/flake.nix#L35

roberth: Not yet.

This command doesn't perform the link check, making it a bit quicker. I'm trying to get the lychee link checker to run in the sandbox, and then checking will be quick and we can settle on that solution.

srid: Some sort of a general tool would be nice, that I can pull from a smaller flake (rather than pull all of flake-parts website flake inputs).

Will merge this for now, but will have to adjust the way CI runs this check in the future.

PR #537, opened 2024-07-01
Fix og:image path to be absolute

TristanCacqueray:
This change resolves the relative note image path to make an absolute url.

For `blog/post.md` with a `img`, the og:image is:
- Before: `siteUrl/media/image.png`
- After:  `siteUrl/blog/media/image.png`

### DIFF
--- a/emanote/src/Emanote/Model/Note.hs
+++ b/emanote/src/Emanote/Model/Note.hs
@@ -30,7 +30,7 @@ import Network.URI.Slug (Slug)
 import Optics.Core ((%), (.~))
 import Optics.TH (makeLenses)
 import Relude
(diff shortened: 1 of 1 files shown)


srid:
cc @shivaraj-bh

srid (commented):
I'm not entirely sure about this change. Been a while since I worked on Emanote, but IIRC link and image URLs are not always considered relative.

In any case, we need to tests before being able to make changes like this.

TristanCacqueray:
> In any case, we need to tests before being able to make changes like this.

You are absolutely right, let me convert this to a draft. There are other papercuts in the `og:description` that I'd like to fix (like wikilink not being rendered properly and the footnote). Is there an example test we could use to validate such fixes?

PR #549, opened 2024-09-08
dual-license files

dezren39:
according to https://github.com/srid/emanote/discussions/545

took a first crack at this, there's a lot of ways one could go about it.


srid:
Can we avoid duplicate content? Just license `emanote/default` while stating the necessary exceptions.

PR #550, opened 2024-09-24
Add ema:note:date value to the template

TristanCacqueray:
This change enables displaying the note date.

### DIFF
--- a/emanote/src/Emanote/View/Template.hs
+++ b/emanote/src/Emanote/View/Template.hs
@@ -1,6 +1,7 @@
 module Emanote.View.Template (emanoteSiteOutput, render) where
 
 import Control.Monad.Logger (MonadLoggerIO)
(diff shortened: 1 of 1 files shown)


srid:
Is it not already possible to do this with `<ema:metadata>`?

See https://github.com/srid/emanote/discussions/131#discussioncomment-1333646

TristanCacqueray:
Unfortunately that doesn't work, it seems like there is a bug for this template. The HTML does get rendered, but the text content is empty.

TristanCacqueray:
Alright, I thought the value would be printed when using `<value var="date">`, but that just set the tag value, so I needed to actually use the tag:

```
<!-- Create a new tag -->
<bind tag="date"><ema:metadata><value var="date" /></ema:metadata></bind>
<!-- Show the date -->
<date />
```

PR #552, opened 2024-09-29
Set note `date` when the path begin with a timestamp

TristanCacqueray:
This change simplifies timeline creation by setting the date metadata from the filename when it begins with `YYYY-MM-DD-`.

### DIFF
--- a/emanote/src/Emanote/Model/Note.hs
+++ b/emanote/src/Emanote/Model/Note.hs
@@ -10,6 +10,7 @@ import Control.Monad.Logger (MonadLogger)
 import Control.Monad.Writer (MonadWriter (tell), WriterT, runWriterT)
 import Data.Aeson qualified as Aeson
 import Data.Aeson.Optics qualified as AO
+import Data.Char (isDigit)
 import Data.Default (Default (def))
 import Data.IxSet.Typed (Indexable (..), IxSet, ixFun, ixList)
 import Data.IxSet.Typed qualified as Ix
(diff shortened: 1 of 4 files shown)


emanote/src/Emanote/Model/Note.hs
```diff
@@ -316,7 +320,19 @@ parseNote scriptingEngine pluginBaseDir r src@(_, fp) s = do
         parseNoteMarkdown scriptingEngine pluginBaseDir fp s
       R.LMLRoute_Org _ -> do
         parseNoteOrg s
-  pure $ mkNoteWith r (Just src) doc meta errs
+  let metaWithDateFromPath = case P.parse dateParser mempty (takeFileName fp) of
+        Left _ -> meta
+        Right date -> SData.modifyAeson (pure "date") (const $ Just $ Aeson.String date) meta
```
srid: If the note already has `date` metadata, I'd think we should use that over the inferred one.

TristanCacqueray: Good point, I updated the PR to preserve existing metadata.

srid (changes requested):
LGTM.

Could you:

- fix the CI failure
- mention this in the docs around [here](https://emanote.srid.ca/guide/yaml-config#special-properties)

image

TristanCacqueray:
Great, I've updated the change.

srid (approved):
👍🏿 I don't have the time to test this locally but I trust that you have done so.

PR #553, opened 2024-09-30
Add support for extra navigation with next/prev links

TristanCacqueray:
This change enables setting ema:note:next and ema:note:prev template to insert extra navigation link in the page from the metadata.

### DIFF
--- a/docs/guide/yaml-config.md
+++ b/docs/guide/yaml-config.md
@@ -22,6 +22,23 @@ Notice how this page's sidebar colorscheme has [changed to green]{.greenery}? Vi
 - `page.image`: The image to use for the page. This is used for the [[ogp]] meta tag `og:image` meta tag. If not specified, the first image in the page is used. Relative URLs are automatically rewritten to absolute URLs if `page.siteUrl` is non-empty.
 
 
+- `next`, `prev`: Lists of wikilinks to be used for custom navigation. When set, the following template is available:
+
+```
+<ema:has:prev>
+  <div class="flex-1 p-4 mt-8 bg-gray-100 rounded">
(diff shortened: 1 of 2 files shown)


TristanCacqueray:
Hello @srid, with this change I'm trying to setup extra navigation links between notes, for example to suggest the next thing to read. I guess this can be done as part of the note, but I'd like to have a consistent theme similar to the backlinks component so this should be displayed via a template. Also, we might be able to automatically add the prev/next value for chronological notes like for a blog.

Does that sounds like a good idea? I wrote this PR quickly to test this feature for my new [video](https://midirus.com/video) feed, which is presently not very usable since you have to go back to the directory to pick the next note.

srid:
Do you have to define `next` and `prev` explicitly for each note? Is it not possible to automatically determine them from a listing (query timeline for example)?

Generally a good idead to expose 'next' and 'prev' note to templates, yes.

srid:
>  we might be able to automatically add the prev/next value for chronological notes like for a blog.

Oh right you stated this too. Yes, something like that would be ideal. Basically like how `mdBook` works (the arrows on bottom of https://omnix.page/om/show.html for example)

TristanCacqueray:
Ok that's good to hear, I'll clean-up this initial change, then I'll look into adding the values automatically.

I already tried to implement a `lookupSiblingNotes :: Model -> R.LMLRoute -> (Maybe R.LMLRoute, Maybe R.LMLRoute)`, but I'm not sure how to use the `IxNote` to get the notes sharing the same parent. Should I just use `allRoutes = Set.fromList $ fmap N._noteRoute $ Ix.toList $ model ^. modelNotes` and find the prev/next from there?

PR #561, opened 2024-10-24
Add container image

TristanCacqueray:
This change enables running emanote in a container.


TristanCacqueray:
Note that I haven't tested the CI action, but that's the script I'm using elsewhere and I hope it will work as-is.

srid (commented):
@TristanCacqueray This is what I see:

image

TristanCacqueray:
That's odd, the script works with the default action runner, here is an example: https://github.com/logjuicer/logjuicer/actions/runs/10422103187/job/28865852855

I guess the XDG_RUNTIME_DIR of your runner is not writable, perhaps try setting it to a valid location for this task?

srid:
> XDG_RUNTIME_DIR

Curious this is a requirement of skopeo? If so [crane](https://github.com/google/go-containerregistry/blob/main/cmd/crane/doc/crane.md) may be worth looking into, as we've got it working with the same kind of runners: https://github.com/nammayatri/nammayatri/blob/acced803e46a801edada92cd1fd5fa769790deb2/.github/workflows/nix.yaml#L113-L127

As a stretch goal only, it'd be good to use https://github.com/nlewo/nix2container over `pkgs.dockerTools`.

TristanCacqueray:
@srid I think skopeo/podman writes the credentials to /run .

Also, once the package is created on GH you'll have to mark at as public so that it can be used freely.

srid:
@TristanCacqueray 

I got the image push working by switching to `crane`. I had to re-open this PR because forks can't push images, it seems. Let's move this discussion to #586

Issue #465, opened 2023-12-04
Obsidian-style callouts


srid:
Spec: https://help.obsidian.md/Editing+and+formatting/Callouts

Discussion: https://github.com/srid/emanote/discussions/307

### Progress

- [x] Basic callouts #466
- [x] [bug] Don't render footnotes inside of callout (they are already rendered in the page)
- [x] Support the rest of callout types
- [x] Nested callouts
- [x] Foldable callouts

Tangential

- [ ] Should we switch to using https://github.com/jgm/commonmark-hs/pull/135 ? cf. https://github.com/srid/emanote/pull/496

srid:
One way to implement this (especially as Pandoc AST is not flexible enough) is by having `heist-extra` generate the same HTML div structure as Obsidian's publish:

https://github.com/srid/heist-extra/blob/54ff970f733dd45b5509d1c4c298927b6241041b/src/Heist/Extra/Splices/Pandoc/Render.hs#L67-L69

<img width="557" alt="image" src="https://github.com/srid/emanote/assets/3998/7c088303-61c4-49be-b1c7-bbce16d0b0ce">

<img width="413" alt="image" src="https://github.com/srid/emanote/assets/3998/57dc3f3b-67c6-4421-aa50-18fcf57c0ae5">


And then style it in CSS separately.

srid:
Actually we could just implement it as a `PandocBlockRenderer` (much like how embed wikilinks, queries, etc. are done)

https://github.com/srid/emanote/blob/ea2dca29a92430572048eb5ad5f018b0e66c42bc/emanote/src/Emanote/Pandoc/Renderer/Query.hs#L25-L41

nramirezuy:
It would be nice to be able to add my own callouts.

Syntax:
```markdown
> [!name] callout title
> callout body
```

[This is so nasty it might as well be AI generated, but works. If someone interested (: ](https://gist.github.com/nramirezuy/3f669efab672530d27d7256497a26d97)

srid:
@nramirezuy That'd be a good feature to add. It would go in this Haskell module,

https://github.com/srid/emanote/blob/master/emanote/src/Emanote/Pandoc/Renderer/Callout.hs

srid:
@nramirezuy What is the difference between:

```markdown
> [!name] callout title
> callout body
```

and

```markdown
> [!note] callout title
> callout body
```

?

[...]

PR #572, opened 2025-02-17
Factor out callout styling in one place

srid:
This will make it easy to work with custom callouts (#465)

(One more PR upcoming to make it actually possible to add custom callouts without modifying emanote)


srid:
cc @nramirezuy

https://github.com/srid/emanote/discussions/307#discussioncomment-12125663

PR #584, opened 2025-07-19
nix: Update all inputs

srid:
Gets us newer GHC.


srid:
Branch name fucked up with ghostty https://github.com/srid/emanote/pull/585

PR #596, opened 2025-07-23
font: Bring back Maven Pro

srid:
We dropped using Maven Pro #345 and then went to default fonts in #574 ... but that wasn't a great decision as it stripped Emanote's visuals of its individuality.

I realized this only upon browing [an old Emanote site](https://themotte.zettel.page/), which looks much better than the then current one.


srid:
https://emanote.srid.ca/fonts

PR #610, opened 2025-09-18
Fix hljs loading

edenworky:
https://github.com/highlightjs/highlight.js/issues/2980


srid (commented):
@edenworky Could you provide some context as to why this is needed when hljs seems to already work, say in https://emanote.srid.ca/syntax-highlighting#example-highlightjs

srid:
Obviated by https://github.com/srid/emanote/pull/624

PR #617, opened 2025-09-28
Add dark-mode support to special pages of the template

git-morin:
Hi 👋 

I really like emanote; here's some changes I found missing from the last feature for dark mode. 

As you can see here (assuming you have dark mode ON) : 

- https://emanote.srid.ca/-/all
- https://emanote.srid.ca/-/tags
- https://emanote.srid.ca/-/tasks
- (also the stork search component)

It looks like the dark-mode feature was not added to the special pages in the template. I've taken the liberty of adding them following what you had done in the main feature code merge.

I've deployed this revision here : https://git-morin.github.io/nixOS/-/all.html (please note most of my current pages are empty or minimal, I just started my journey of moving my notes to obsidian/emanote format and doing a sort of rewrite in the process)

Thanks for this amazing work, et vive le Québec ⚜️💙 !


srid (changes requested):
Hi, great to meet a fellow Quebecer!

All looks good on my browser. But the stork changes seem redundant, as we would be better off directly using one of the upstream dark themes from here:
https://github.com/jameslittle230/stork/tree/master/themes

Perhaps we can switch to the `edible` theme as it provides dark variant.

Image

You can put those CSS files under https://github.com/srid/emanote/tree/master/emanote/default/_emanote-static/stork and reference them.

git-morin:
I've added `edible` with its dark variant, and the needed code in the stork component to swap between based on the `prefers-color-scheme`. I've also removed flat.css since it was not referenced (but let me know if I can fix further or you would like to keep it anyways)

Looks like this:

image


Cheers 🥂

srid:
Looks great. Sorry I didn't get any email notification about your comment for whatever reason; just saw this.

srid (approved):

PR #618, opened 2025-10-10
mermaid: add elk layout algorithm

tropf:
Dear maintainer,

I'd like to make the `elk` layout option available for mermaid diagrams.
In complex/convoluted diagrams, elk can generate better-readable diagrams.

This is achieved by loading an additional dependency and initializing it, which I copied from [the `mermaid-js/layout-elk` npm package](https://www.npmjs.com/package/@mermaid-js/layout-elk).

I have also added a short example in the emanote docs.

See also: [mermaid layout documentation](https://mermaid.js.org/config/layouts.html)

Screenshot from docs:  
image

### DIFF
--- a/docs/tips/js/mermaid.md
+++ b/docs/tips/js/mermaid.md
@@ -148,3 +148,47 @@ gantt
     Add gantt diagram to demo page      :20h
     Add another diagram to demo page    :48h
 ```
+
+### Layout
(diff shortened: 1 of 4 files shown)


srid (commented):
Thanks!

(Will test prior to merge)

PR #624, opened 2025-12-19
Add built-in static syntax highlighting using skylighting

### LINKED ISSUE #337: Static syntax highlighting
Use Pandoc to syntax highlight code blocks statically so we do not need to rely on client-side syntax highlighters (like highlight.js).

It should be possible to disable this feature in [Yaml configuration](https://emanote.srid.ca/demo/yaml-config) for users who'd rather use a client-side syntax highlighter.

srid:
Replaces client-side JavaScript syntax highlighters (highlight.js, prism.js) with static highlighting using skylighting, the same library used by Pandoc. 

Key improvements:
- Code blocks are now highlighted at build time using skylighting (140+ languages supported)
- No JavaScript required, resulting in faster page loads and better compatibility
- Works seamlessly with live preview mode
- Both light and dark mode themes included
- Configurable via `syntaxHighlighting` option (default: true)

The documentation has been updated to reflect the new approach, and legacy JavaScript snippets have been removed from example configurations.

Resolves #337
Uses https://github.com/srid/heist-extra/pull/10


srid:
https://emanote.srid.ca/syntax-highlighting

Issue #520, opened 2024-03-05
Render TOC without JavaScript


srid:
> As for `highlightToc`, do we really need JavaScript? I see a NoJS solution here: https://codepen.io/davorsuljic/embed/vYgqKWe? (via the comments from https://css-tricks.com/sticky-table-of-contents-with-scrolling-active-states/ )

_Originally posted by @srid in https://github.com/srid/emanote/issues/504#issuecomment-1937783917_

TristanCacqueray:
I get the appeal, but this css is really advanced, not sure what's the benefit at that point. If I understand correctly, each heading needs it's own css to match the relevant doc section.

PR #636, opened 2026-04-21
UI pass: typography, theme toggle, backlink cards, TOC hierarchy, self-hosted fonts, MCP shell

srid:
## Summary

A top-to-bottom visual refresh of the default Emanote theme, arrived at through iterative feedback against the live dev server. The goal was to land somewhere *colourful and fun but not jarring* — evocative of [neuron.zettel.page](https://neuron.zettel.page/)'s character, suitable for Zettelkasten notes as much as tech docs or personal pages — without making the theme feel like a template everyone else already uses.

### What changed

**Typography.** Swapped the old Source Serif stack for **Lora** (prose), **Space Grotesk** (UI chrome + headings), and **Space Mono** (code). All three are self-hosted under `_emanote-static/fonts/` as woff2 + a generated `fonts.css` with relative URLs, so generated static sites no longer phone home to `fonts.gstatic.com`.

**Theme toggle.** A persistent dark/light mode toggle lives in the sidebar (and, critically, in the no-sidebar "neuron-style" layout too, via the top-right button row). State is kept in `localStorage`; an early-load script in `base.tpl` applies the `.dark` class before first paint to avoid a flash of wrong theme. Everything previously keyed off `prefers-color-scheme` — `skylighting.css`, the stork search dialog — was rewired to the manual toggle via `.dark` class selectors (and a tiny `MutationObserver` in the stork wrapper for live swap).

**Note title** is back to the Emanote brand treatment: `bg-primary-600`, white, centered, `rounded-2xl`. That look *is* the Emanote brand; an earlier iteration stripped it and it immediately felt generic.

**Backlinks.** Re-done as a 2-column card grid with a thin primary left-accent (`border-l-2`) and subtle shadow, no outer panel bg/border. The "Links to this page" block now picks up the sans font via `#backlinks` being added to the UI-chrome selector.

**Links.** Wikilinks get a subtle primary-50/primary-950 **hover background** (no underline); external links switch to hover-only underline. The wikilink background uses `-mx-1 px-1 rounded` so text doesn't reflow on hover.

**TOC.** Replaced the old `window.onscroll` scroll-spy with an `IntersectionObserver` (#520's long-standing TODO). On top of that: depth-based visual hierarchy — each nested `<ul>` level in the "On this page" list steps down font-size and colour, so a 30-heading page stays scannable at a glance. Active-item styling now uses `var(--color-primary-*)` so per-site theme overrides flow through cleanly.

**Sidebar.** Removed the tag-index and expand-tree shortcut icons — search and theme toggle only. Folgezettel tree depth rails (`#sidebar .pl-2 .pl-2 { border-left: ... }`) keep the nested structure legible.

**Smaller polishes.** Callouts use `color-mix(in srgb, ${color} 8%, transparent)` for tint backgrounds (no more naive `${color}10`). External-link glyphs are drawn with `mask-image` + `currentColor` so they inherit link colour in both themes. `kbd` restyled with theme variables. Task-list items use `:has(> svg.--ema-checkbox)` to swap the disc bullet for a flex-aligned checkbox. Footnotes re-done as a semantic `<aside> + <ol>` instead of the old `scale-x-90` squish. Nested-list vertical rhythm fixed. Mermaid diagrams re-render on theme toggle.

**Infra/MCP.** Added `{{nix_shell}}` prefixing in `justfile` so `just run` works outside a Nix shell, plus a thin `just mcp-chrome-devtools` recipe invoked from `.mcp.json`. A new `nix/chrome-devtools/shell.nix` bundles Chromium + pins `chrome-devtools-mcp@0.21.0` (adapted from [juspay/kolu#650](https://github.com/juspay/kolu/pull/650)) so the chrome-devtools MCP server has a headless browser to drive.

**Docs.** `docs/guide/html-template/fonts.md` updated to reflect the new default stack.

### Notable tradeoffs

- Self-hosted fonts add ~612 KB to the static layer (29 woff2 blobs across Lora + Space Grotesk + Space Mono subsets). The offline-capable generated site is worth the bytes; browsers cache the blobs per-origin anyway.
- Dropping the tag/expand-tree shortcuts from the sidebar is a deliberate simplification — the pages are still reachable via breadcrumbs and the full URL.

## Test plan

- [ ] Verify both layouts (with-sidebar and neuron-style no-sidebar) render correctly in light + dark mode.
- [ ] Exercise the theme toggle from the sidebar and from the neuron-layout top-right button; confirm persistence across reloads.
- [ ] Open the stork search dialog in both themes; confirm input, results list, and highlight contrast are correct.
- [ ] View a code block in both themes; confirm syntax highlighting follows the manual toggle (not just OS preference).
- [ ] Browse a long-outlined page (e.g. `/yaml-config`) and confirm the TOC shows depth-based hierarchy and the active-item tracker follows scroll.
- [ ] Confirm `cabal build all` and `cabal test all` succeed.
- [ ] Confirm generated static site produces no `fonts.googleapis.com` network requests.


nix/chrome-devtools/shell.nix
```diff
@@ -0,0 +1,23 @@
...
+# with Chromium bundled, so `just mcp-chrome-devtools` works on machines
+# without a system Chrome binary.
+#
+# Adapted from juspay/kolu#650: the upstream flake packages only the MCP
+# server and defers Chrome provisioning to the consumer. Emanote is the
+# consumer here, so Chromium is pinned and wired into --executable-path.
+{ pkgs ? import
+    (fetchTarball {
+      url = "https://github.com/NixOS/nixpkgs/archive/nixos-unstable.tar.gz";
+    })
```
srid: Please use the same nixpkgs as toplevel `flake.lock`

srid: Done in f37d72af — shell.nix now reads flake.lock directly via builtins.fetchTree (owner/repo/rev/narHash) instead of pinning nixos-unstable separately. One pin to maintain.

PR #639, opened 2026-04-21
Static math on by default: LaTeX → MathML at build time

### LINKED ISSUE #626: Render math as native MathML via Pandoc (offline support)
## Problem

Currently, math rendering requires client-side JavaScript loaded from CDNs:

- **MathJax**: `cdn.jsdelivr.net/npm/mathjax@3`
- **KaTeX**: `cdn.jsdelivr.net/npm/katex@0.16.9` (CSS + JS + auto-render)

These are opt-in via `<snippet var="js.mathjax" />` or `<snippet var="js.katex" />` in `page.headHtml`. Generated sites cannot render math when viewed offline.

## Proposed Solution

Use Pandoc's built-in `--mathml` support (`writerHTMLMathMethod = MathML`) to render TeX math as **native MathML** elements at build time.

### Implementation

1. Set `writerHTMLMathMethod = MathML` in the Pandoc writer configuration
2. Pandoc will emit `<math>` MathML elements directly in the HTML output
3. No client-side JavaScript needed — browsers render MathML natively
4. Remove or deprecate `js.mathjax` and `js.katex` snippets (or keep as opt-in fallback)
5. Optionally add a small CSS snippet for MathML styling

### Browser Support

Native MathML is supported in:

- Firefox (long-standing support)
- Safari (long-standing support)
- **Chrome 109+** (January 2023)

This covers the vast majority of current browsers.

### Benefits

- Generated site works fully offline
- No client-side JavaScript needed for math
- Faster page loads
- Math is part of the HTML document structure

### Tradeoffs

- Very old browsers (Chrome < 109) won't render math correctly
- MathML rendering may look slightly different from MathJax/KaTeX

srid:
**`$...$` / `$$...$$` now render to MathML at build time**, so a minimal Emanote site ships zero math JavaScript by default. Uses the new `RenderFeatures` API from [heist-extra#11](https://github.com/srid/heist-extra/pull/11), which swaps a grab-bag of per-feature `Bool`s on `RenderCtx` for a single record with sum-typed backends (`CodeBackend`, `MathBackend`). Closes #626.

Modern browsers (Firefox, Safari, Chrome ≥109) render MathML natively, so the default flips to `emanote.staticMath: true`. The `js.katex` snippet is **removed** from the default config — KaTeX is a minor enough alternative path that shipping a pinned CDN loader with SRI hashes that diverge from the KaTeX docs was doing readers a disservice. The MathJax snippet stays because it's the documented fallback. *Existing sites referencing `<snippet var="js.katex" />` will need to paste the loader directly into `page.headHtml` — see the rewritten `docs/tips/js/math.md` for the exact lines.*

An e2e smoke scenario asserts every build produces a `<math display="inline">` and a `<math display="block">` on the math fixture page, and that no KaTeX asset is referenced on a default-config page.

### Try it locally

```sh
nix run github:srid/emanote/static-math -- run -L docs
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ apm_modules/
 
 /.worktrees
 /.do-results.json
+/.claude/scheduled_tasks.lock
--- a/docs/guide/html-template/external-links.md
+++ b/docs/guide/html-template/external-links.md
@@ -1,8 +1,5 @@
 ---
 slug: external-links
-page:
-  headHtml: |
-    <snippet var="js.mathjax" />
 ---
 
 # External links
--- a/docs/guide/orgmode.org
+++ b/docs/guide/orgmode.org
@@ -15,7 +15,7 @@ Here is a handpicked selection of syntatic features of Org Mode as particularly 
 
 *** Code blocks 
 
-See [[file:../tips/js/syntax-highlighting.md][Syntax Highlighting]] for general information.
+See [[file:../tips/syntax-highlighting.md][Syntax Highlighting]] for general information.
 
 #+NAME: factorial
 #+BEGIN_SRC haskell :results silent :exports code :var n=0
--- a/docs/guide/orgmode.yaml
+++ b/docs/guide/orgmode.yaml
@@ -1,3 +1,2 @@
-page:
-  headHtml: |
-    <snippet var="js.mathjax" />
+slug: orgmode
+
--- a/docs/tips/js/math.md
+++ b/docs/tips/js/math.md
@@ -9,7 +9,7 @@ Emanote renders `$...$` and `$$...$$` to **MathML at build time** by default via
 ### Demo
 
 When $a \ne 0$, there are two solutions to $ax^2 + bx + c = 0$ and they are
-$$x = {-b \pm \sqrt{b^2-4ac} \over 2a}.$$
+$$x = \frac{-b \pm \sqrt{b^2 - 4ac}}{2a}.$$
 
 ## Opting out
 
@@ -39,8 +39,15 @@ Paste the KaTeX loader directly into `page.headHtml` — Emanote's default confi
 ```yaml
 page:
   headHtml: |
-    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/katex@0.16.9/dist/katex.min.css" crossorigin="anonymous">
-    <script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.9/dist/katex.min.js" crossorigin="anonymous"></script>
-    <script defer src="https://cdn.jsdelivr.net/npm/katex@0.16.9/dist/contrib/auto-render.min.js" crossorigin="anonymous" onload="renderMathInElement(document.body);"></script>
+    <link rel="stylesheet"
+          href="https://cdn.jsdelivr.net/npm/katex@0.16.9/dist/katex.min.css"
+          crossorigin="anonymous">
+    <script defer
+            src="https://cdn.jsdelivr.net/npm/katex@0.16.9/dist/katex.min.js"
+            crossorigin="anonymous"></script>
+    <script defer
+            src="https://cdn.jsdelivr.net/npm/katex@0.16.9/dist/contrib/auto-render.min.js"
+            crossorigin="anonymous"
+            onload="renderMathInElement(document.body);"></script>
 ```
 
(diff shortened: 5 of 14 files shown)


srid:

| # | Lens   | Finding                                                                          | Disposition      |
|---|--------|----------------------------------------------------------------------------------|------------------|
| 1 | Lowy   | flake.nix branch pin is an untracked time-bomb                                   | Fixed in this PR |
| 2 | Lowy   | Config reading in View/Common.hs groups by feature, not by volatility axis       | Deferred         |
| 3 | Lowy   | Default-on + KaTeX snippet removal are two volatility axes crammed in one release | Deferred         |

### Hickey rationale

No findings. The `renderFeatures` builder's two `if/then/else` arms looked like copy-paste-with-variation, but the agent judged extraction premature at two features. The `Bool → RenderFeatures` signature swap in `Renderer.hs` is a strict improvement over the prior lone-Bool shape.

(The `/simplify` pass flagged the `if/then/else → bool` swap separately; fixed in `de311c08`.)

### Lowy rationale

**L1 (fixed)** — the `flake.nix` pin to `github:srid/heist-extra/static-math` is a moving target. A TODO comment now anchors the intent so a rebuild months from now either flags drift or prompts the replacement with a tagged ref. Commit `551eabb8`.

**L2 (deferred)** — the `renderFeatures` builder lives in `View/Common.hs` and hand-rolls YAML→sum-type mapping for both features in one block. At a third backend this pattern starts to smell, and per-page override (as opposed to site-wide) has no home here. The right seam is `readRenderingConfig :: Aeson.Value → RenderingConfig` + `toRenderFeatures :: RenderingConfig → Splices.RenderFeatures`, but with only two features the abstraction is speculative. Revisit when the third backend lands or when per-page override is a concrete need.

**L3 (deferred)** — shipping `staticMath: true` *and* removing the `js.katex` snippet in the same release bundles "new default-on feature" with "break a documented extension point." The volatility-correct split would be: flip the default in this PR with the KaTeX snippet still present (maybe deprecation-tagged), then remove the snippet in a follow-up. The user explicitly asked for both in one release — accepting the coupling for now and leaning on docs/CHANGELOG to carry the migration.

srid:
image

srid:

| research | ✓ | 30s | mapped integration points; one caller |
| branch | ✓ | 0s | static-math off origin/master |
| implement | ✓ | 1m 18s | RenderFeatures threaded through Renderer+Common; staticMath config added |
| check | ✓ | 1m 58s | `cabal build all` clean |
| docs | ✓ | 24s | rewrote math.md; CHANGELOG bullet |
| fmt | ✓ | 34s | fourmolu clean |
| commit | ✓ | 1m 1s | primary feature commit |
| hickey+lowy | ✓ | 2m 47s | 1 fixed (flake-pin TODO); 2 deferred (per-feature extraction / default-on coupling) |
| police | ✓ | 5m 44s | elegance: if/then/else → bool |
| test | ✓ | 1m 2s | e2e: inline + display MathML + no-KaTeX scenarios |
| create-pr | ✓ | 1m 7s | this PR |
| **ci** | ✓ | **14m 58s** | `vira ci` signed off on HEAD |
| **Total** | | **32m 40s** | |

**Slowest step**: `ci` (14m 58s) — vira runs `nix build` across both `aarch64-darwin` and `x86_64-linux` with caching disabled, so two serial emanote-binary builds carry almost all of that.

### Optimization suggestions

- **ci dominated** at 46% of total. Running `vira ci --only-build` (skips cache+signoff stages) against HEAD would cut the first build in half; pairing that with a `vira ci` at the end (for the signoff) keeps green-ticks without paying the double cost between commits.
- **Two CI runs** this PR (one after initial push, one after the heist-extra unpin). On a branch that's likely to accumulate follow-up commits, `--from ci-only` avoids replaying hickey/lowy/police every time.
- **police's 5m 44s** came almost entirely from the `/simplify` lens catching the `if/then/else → bool` swap. Pre-running `/simplify` locally before `/do` would let the police step land in one pass.
- **Dev-server iteration** caught two real bugs (`\over` unsupported by texmath; long KaTeX lines overflowing the `<pre>`) and a third (broken `tips/js/syntax-highlighting` link in orgmode.org). Those wouldn't have surfaced from CI alone — worth keeping the "spin up dev server, click through affected pages" loop after any static-rendering change.

Workflow completed at 2026-04-21T18:51:00Z.

PR #640, opened 2026-04-21
Namespace footnote ids per embedded note

### LINKED ISSUE #360: Footnotes and embedding: duplicate HTML element ids

srid:
**A note containing a footnote that embeds another note — which also has a footnote — rendered two `id="fn1"` and two `id="fnref1"` on the same page.** Valid HTML requires unique ids, and `document.getElementById("fn1")` on such a page is non-deterministic. Both the parent and the embed go through heist-extra's `pandocSplice`, which numbers footnotes from 1 per document with no shared namespace — so without a knob in the library, the consumer can't fix this. [srid/heist-extra#12](https://github.com/srid/heist-extra/pull/12) adds that knob (`RenderCtx.idPrefix`); this PR consumes it.

`Emanote.Pandoc.Renderer.Embed.embedResourceRoute` now builds a per-embedded-note prefix from the embedded note's URL (non-alphanumerics sanitised to `-`) and sets it on the `RenderCtx` passed into the embed's `pandocSplice` call. The template `pandoc.tpl` switches `id="fn${footnote:idx}"` to `id="fn${footnote:id}"` (and the mirrored `fnref` and `href` fragments); *the visible `<footnote:idx />` digit stays unprefixed so readers keep seeing ¹ ² ³*. Top-level rendering uses `idPrefix = ""`, so non-embedded pages are byte-for-byte unchanged.

The e2e suite gains a fixture (outer note with footnote, embedding an inner note with footnote) and a scenario that asserts every id on the rendered page is unique and at least one `fnref` id is present — the second assertion is narrower than `fn*` specifically so a heading slug like `function-call` can't masquerade as a footnote ref.

> **Hitchhiker**: `docs/tips/js/math.md` moves to `docs/tips/math.md`. Math is statically rendered via `texmath` now — it's not a JS behaviour, so it doesn't belong under `tips/js/`. The slug (`/math`) is unchanged and the legacy `/tips/js/math /math 301` redirect stays in place.

Depends on [srid/heist-extra#12](https://github.com/srid/heist-extra/pull/12) — `flake.nix` tracks the `footnote-id-prefix` branch until that merges.

Closes #360

### Try it locally

```sh
nix run github:srid/emanote/warm-nurse -- -L ./docs run
```


srid:

| # | Lens   | Finding                                               | Disposition           |
|---|--------|-------------------------------------------------------|-----------------------|
| 1 | Hickey | `[id^="fn"]` locator too broad (heading slug false-positive) | **Fixed** (ff0631ea) |
| 2 | Hickey | `sanitizeForId` name is generic for a URL-only helper | Defer                  |
| 3 | Hickey | Empty-URL edge case in prefix construction            | No-op — `SR.siteRouteUrl` invariant |
| 4 | Hickey | `js/math.md` rename hitchhiking                       | Accept — user-requested |
| 5 | Lowy   | Prefix policy inlined at call site, no named helper   | **Fixed** (42e1313e)  |
| 6 | Lowy   | heist-extra/emanote boundary tracks volatility cleanly | No-op                 |
| 7 | Lowy   | `sanitizeForId` half-measure (private but general-purpose) | Defer — promote when a 2nd consumer appears |
| 8 | Lowy   | flake input pinned to branch, not commit              | Accept — branch ref per maintainer preference (318dfe95) |
| 9 | Lowy   | math.md relocation is a separate axis of change       | Accept — user-requested, low-risk |

### Hickey rationale

`sanitizeForId`, the embed ctx update, and the template's two-splice split (`footnote:id` for HTML ids, `footnote:idx` for the visible digit) are principled — the two splices name different things, so they're not incidental duplication. Only one genuine fact-check finding landed: the `[id^="fn"]` locator in the new e2e step would have let a future heading slug like `final-notes` masquerade as a footnote ref. Tightened to `[id^="fnref"]` in ff0631ea. The empty-URL guard was rejected per CLAUDE.md (internal invariant from `SR.siteRouteUrl`).

### Lowy rationale

The heist-extra / emanote boundary tracks volatility correctly: heist-extra owns the mechanism (`RenderCtx.idPrefix` + `footnote:id` splice) and stays policy-neutral, emanote owns the policy (what value the prefix takes). If the prefix scheme changes, only emanote's `Embed.hs` moves. Extracting `computeEmbedIdPrefix` names that policy decision so swaps stay mechanical. The flake-input pin-vs-branch call was flipped back to branch ref per maintainer preference — `flake.lock` still captures the commit for reproducibility.

srid:

| research | ✓ | 4m 2s | Two-repo plan confirmed (heist-extra mechanism + emanote consumer) |
| branch | ✓ | 10s | emanote on `warm-nurse`, heist-extra on `footnote-id-prefix` |
| implement | ✓ | 4m 4s | idPrefix plumbed through, pandoc.tpl updated, e2e fixture + scenario, `docs/tips/js/math.md` → `docs/tips/math.md` |
| check | ✓ | 1m 13s | `cabal build all` exits 0 |
| docs | ✓ | 16s | CHANGELOG + one orgmode link + `default/index.yaml` comment |
| fmt | ✓ | 21s | cabal-fmt, fourmolu, hlint, nixpkgs-fmt all pass |
| commit | ✓ | 29s | primary commits pushed to both branches |
| hickey+lowy | ✓ | 4m 2s | 4 parallel sub-agents; 2 Fix-now on emanote, rest Defer/No-op |
| police | ✓ | 11m 50s | rules clean; fact-check tightened the footnote locator; elegance trimmed haddocks, hoisted `show idx`, switched `RenderCtx` to named-field syntax |
| test | ✓ | 46s | `cabal test all`: 23 examples, 0 failures (Playwright e2e runs in GitHub Actions) |
| create-pr | ✓ | 4m 39s | srid/heist-extra#12 + this PR; hickey/lowy analysis posted on each; inline review on heist-extra (version revert) resolved |
| ci | ✓ | 7m 57s | `vira ci` signoffs green on aarch64-darwin + x86_64-linux for both repos (one iteration; htmlproofer caught a cross-page href regression and the template split was tightened) |
| **Total** | | **40m 23s** | |

**Slowest step**: `police` (11m 50s)

### Optimization suggestions

- **police dominates at 11m 50s** — the three-pass design (rules / fact-check / elegance via `/simplify` parallel fan-out) is doing real work, but the main cost here was the elegance pass mixing real findings (trim haddock, named-field syntax) with a commit I had to rewrite after an `--amend --no-edit` mangle. For next time: do `git commit --amend --no-edit` only on commits with no `-m` override intended, and split "apply many small elegance findings" into one batch pre-commit if per-finding commits are the convention.
- **ci took 7m 57s over two iterations** — the first iteration caught a `<base href='/'>`-interacting regression in the new fragment-only hrefs. Running a local `nix build '.#docs'` (which also invokes htmlproofer via the `linkCheck` derivation) right after any template change that touches hrefs would catch this in ~30s instead of a full `vira ci` round.
- **research was 4m 2s** — most of that was delegated to an `Explore` sub-agent tracing two repos. For tightly-scoped two-repo fixes, the main context could do the trace directly in ~1m since the file count is small.
- **hickey+lowy at 4m 2s** ran 4 parallel sub-agents. Well-amortised — reviewing both PRs at once was the right structure. No change.

Workflow completed at 2026-04-21T19:52Z.

srid:
Closing in favour of the **container-scoped** approach tracked in #641 — drop footnotes entirely on the embed side, keep callout footnotes inside the callout body. The prefix mechanism here *works* (both vira ci runs green) but produces two "Footnotes" sections on the rendered `/embed` page with duplicated `¹` numbering, which is the wrong UX resolution.

The entire `idPrefix` / `computeEmbedIdPrefix` / `footnote:id` mechanism goes away in the next attempt — the embed's `B.Note` inlines get stripped pre-render, so there's no `<sup>` and no trailing list from the embed at all. Callouts get their own scoped list *inside* the callout block (reader-intuitive — Obsidian-style), and the outer `gatherFootnotes` learns to skip callout subtrees.

E2E fixtures and the hickey/lowy analysis from this PR remain useful context; the replacement PR will cite this one as prior art. See #641 for the plan.

PR #642, opened 2026-04-22
Popup-only footnotes with print fallback

### LINKED ISSUE #110: Better footnotes UX
Currently, footnotes are rendered very basically. See a demo here: https://note.ema.srid.ca/demo/markdown

We could spice things up here. For inspiration, see https://jamesclear.com/first-principles

Could use JavaScript for this work. Should be [accessible](http://kb.daisy.org/publishing/docs/html/roles.html), and work on mobile.

### LINKED ISSUE #360: Footnotes and embedding: duplicate HTML element ids

srid:
**Footnote refs open a popup — that's the only on-screen UI.** No `#fn1` anchor, no visible bottom list, no scroll-away. The `<sup>` is the click target (role=button, Enter/Space activates). Desktop gets a floating card; mobile gets a bottom-sheet. For print, the hidden `<aside>` flips to `print:block` with a decimal list so paper copies still carry the cited bodies.

The popup binds to a `data-footnote-*` contract (ref, id, list, scope), not Pandoc class names. Scope routing handles callouts and embedded notes as their own footnote boundaries — each one can legitimately carry a `fn1`.

Native HTML Popover API; no dep. Popover chrome styled entirely via inline Tailwind classes applied in JS; the only CSS rule that survives is `::backdrop { background: transparent }` (Tailwind has no variant for that pseudo-element).

### Verification

- Chrome-mcp real-click walk on `docs/guide/markdown/embed.html` — all refs (parent / callout / embed) route to correct bodies, `scrollY` stays 0, no hash.
- e2e (cucumber + playwright) covers all three popup paths plus a new print-mode scenario (`page.emulateMedia({ media: "print" })` → at least one aside visible, body content present).
- Vira CI green on `2a7c9238`.

### Try it locally

```sh
nix run github:srid/emanote/footnote-popups -- -L ./docs run
```

### DIFF
--- a/docs/guide/markdown/embed.md
+++ b/docs/guide/markdown/embed.md
@@ -6,12 +6,11 @@ date: 2022-08-03
 
 # Embedding
 
-You can embed files, using `![[..]]` - a syntax inspired by [Obsidian](https://help.obsidian.md/Linking+notes+and+files/Embedding+files). The HTML can be fully customized for each embed type[^parent].
+You can embed files, using `![[..]]` - a syntax inspired by [Obsidian](https://help.obsidian.md/Linking+notes+and+files/Embedding+files). The HTML can be fully customized for each embed types.
 
 > [!warning] 
 > The embed wiki-link syntax must appear on a paragraph of its own, with no other text added next to it.[^blk] Recursive embeds are supported.
 
-[^parent]: A page-level footnote, rendered into the bottom list and also shown in the click-anywhere popup. The popup is a progressive enhancement — the bottom list stays usable when JS is off or on print.
 [^blk]: This constraint is necessary to ensure that the HTML generated remains valid. Embedded content use block elements, which cannot be embedded inside inline nodes.
 
 ## Notes
--- a/docs/guide/markdown/embed.md
+++ b/docs/guide/markdown/embed.md
@@ -6,11 +6,12 @@ date: 2022-08-03
 
 # Embedding
 
-You can embed files, using `![[..]]` - a syntax inspired by [Obsidian](https://help.obsidian.md/Linking+notes+and+files/Embedding+files). The HTML can be fully customized for each embed types.
+You can embed files, using `![[..]]` - a syntax inspired by [Obsidian](https://help.obsidian.md/Linking+notes+and+files/Embedding+files). The HTML can be fully customized for each embed type[^parent].
 
 > [!warning] 
 > The embed wiki-link syntax must appear on a paragraph of its own, with no other text added next to it.[^blk] Recursive embeds are supported.
 
+[^parent]: A page-level footnote, rendered into the bottom list and also shown in the click-anywhere popup. The popup is a progressive enhancement — the bottom list stays usable when JS is off or on print.
 [^blk]: This constraint is necessary to ensure that the HTML generated remains valid. Embedded content use block elements, which cannot be embedded inside inline nodes.
 
 ## Notes
--- a/emanote/default/templates/components/footnote-popup.tpl
+++ b/emanote/default/templates/components/footnote-popup.tpl
@@ -176,11 +176,6 @@
       var body = popover.firstElementChild;
       body.textContent = '';
       body.appendChild(cloneContent(target));
-      if (currentRef && currentRef !== ref) {
-        currentRef.classList.remove('emanote-footnote-active');
-      }
-      currentRef = ref;
-      ref.classList.add('emanote-footnote-active');
       // hidePopover throws InvalidStateError if the popover is already
       // closed. That's the expected case on first open; the throw carries
       // no signal we'd act on, so swallow it.
@@ -192,6 +187,13 @@
         console.warn('[emanote] footnote popover showPopover failed', err);
         return;
       }
+      // Only dirty the active-state after show succeeds — otherwise a
+      // failed show leaves a stale highlight on the ref until next click.
+      if (currentRef && currentRef !== ref) {
+        currentRef.classList.remove('emanote-footnote-active');
+      }
+      currentRef = ref;
+      ref.classList.add('emanote-footnote-active');
       // Defer to next frame so popover width reflects the content just
       // inserted — measuring immediately after showPopover() can center
       // on the previous frame's width.
--- a/emanote/default/templates/components/footnote-popup.tpl
+++ b/emanote/default/templates/components/footnote-popup.tpl
@@ -41,12 +41,12 @@
   .emanote-footnote-popup-body > :last-child { margin-bottom: 0; }
   .emanote-footnote-popup-body p { margin-bottom: 0.5rem; }
 
-  sup.footnote-ref.emanote-footnote-active a {
+  sup[data-footnote-ref].emanote-footnote-active a {
     background-color: var(--color-primary-100);
     border-radius: 3px;
     padding: 0 0.2em;
   }
-  .dark sup.footnote-ref.emanote-footnote-active a {
+  .dark sup[data-footnote-ref].emanote-footnote-active a {
     background-color: var(--color-primary-900);
   }
 
--- a/emanote/default/templates/components/footnote-popup.tpl
+++ b/emanote/default/templates/components/footnote-popup.tpl
@@ -89,6 +89,10 @@
 
     var popoverEl = null;
     var currentRef = null;
+    // Pandoc emits exactly one top-level <aside data-footnote-list> per
+    // page. Live-server reloads the whole document on source changes, so
+    // caching across clicks is safe — no SPA mutations to worry about.
+    var topLevelAside = null;
 
     function ensurePopover() {
       if (popoverEl) return popoverEl;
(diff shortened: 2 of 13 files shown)


srid:

| # | Lens   | Finding                                                      | Disposition      |
|---|--------|--------------------------------------------------------------|------------------|
| 1 | Hickey | Mobile breakpoint duplicated in CSS and JS                   | Fixed in this PR |
| 2 | Hickey | `popoverEl` + `currentRef` two-variable state                | No-op            |
| 3 | Hickey | Stale `getBoundingClientRect` on first click                 | Fixed in this PR |
| 4 | Hickey | Silent `showPopover()` failure                               | Fixed in this PR |
| 5 | Hickey | `cloneContent` stripped by Pandoc class, not data attribute  | Fixed in this PR |
| 6 | Hickey | Unused `vh` variable in `positionPopover`                    | Fixed in this PR |
| 7 | Lowy   | Data-attribute contract as the Heist↔popup seam              | No-op (correct)  |
| 8 | Lowy   | `<style>` + `<script>` colocated in one `.tpl`               | No-op (correct)  |
| 9 | Lowy   | Runtime `closest()` vs compile-time scope-id prefix          | No-op (correct)  |
| 10 | Lowy  | Active-state CSS uses Pandoc class instead of data attribute | Fixed in this PR |

### Hickey rationale

Five things live cleanly in `footnote-popup.tpl`: the data-attribute contract, scope router, popover lifecycle, viewport-aware positioning, and the Popover-API guard. None mix into one another at the module level.

The duplicated mobile breakpoint (CSS `@media (max-width: 640px)` and JS `matchMedia('(max-width: 640px)')`) is two representations of the same fact with nothing enforcing sync. The fix caches a single `MediaQueryList` at module top and documents the coupling. The first-click off-center bug was `getBoundingClientRect` firing before browser layout after `showPopover()` inserted new content; deferring `positionPopover` to `requestAnimationFrame` closes it. The `vh` dead code, the silent `showPopover()` failure, and the `a.footnote-backref` class-name leak in `cloneContent` were all quick to fix — and the backref fix completes the data-attribute abstraction the PR explicitly set out to establish.

The one no-op — `popoverEl` / `currentRef` as a two-variable state machine — becomes risky only under partial-page updates, which Emanote's live-server reload model doesn't produce.

### Lowy rationale

The data-attribute seam is the right volatility boundary: the markup layer emits semantic intent; the presentation layer reads it. Replacing click-to-open with hover-to-open or a sidebar panel touches only `footnote-popup.tpl`; the four emitters remain unchanged. Colocating style + script in one `.tpl` is also right — their axes of change are identical (both evolve when the popup presentation evolves, both are removed together if the feature is removed). A runtime `closest()` walk for scope routing beats pushing per-scope unique-id prefixes up into heist-extra: embedding-depth rules are an Emanote-level concern, not a generic Heist concern.

The one real coupling leak Lowy flagged (beyond the Hickey-caught backref strip) was `sup.footnote-ref.emanote-footnote-active a` in the active-state CSS — a Pandoc class mixed with an Emanote-owned class. Tightened to `sup[data-footnote-ref].emanote-footnote-active a` so the whole contract stays within Emanote's own attribute vocabulary.

srid:

| research | ✓ | 0s | Mapped footnote rendering (pandoc.tpl:70-84), heist-extra splice API (Footnotes.hs:42-61), embed/callout templates, e2e structure. Identified embed-scope collision and refined attribute contract. |
| branch | ✓ | 2s | Created branch footnote-popups off origin/master. |
| implement | ✓ | 3m 2s | Added data-footnote-* attrs, new components/footnote-popup.tpl with scope-aware lookup + Popover-API open + mobile bottom-sheet. e2e fixtures + 3 scenarios + 3 steps. [^parent] footnote to docs/guide/markdown/embed.md for chrome-mcp check. |
| check | ✓ | 6s | cabal build all: Up to date. |
| docs | ✓ | 17s | Documented popup behavior in docs/guide/markdown.md. |
| fmt | ✓ | 11s | pre-commit clean: cabal-fmt, fourmolu, hlint, nixpkgs-fmt. |
| commit | ✓ | 20s | Pushed feature branch with primary feature commit. |
| hickey+lowy | ✓ | 5m 26s | Hickey: 6 findings (5 Fix, 1 No-op). Lowy: 5 seams (2 Fix, 3 No-op). 6 follow-up commits. |
| police | ✓ | 4m 34s | Rules: 2 fixes. Fact-check: 1 fix. Elegance: 5 refactors. 8 commits. |
| test | ✓ | 7m 49s | Chrome-mcp walked 4 refs on /embed.html: parent/callout/embed scopes all routed correctly. Mobile bottom-sheet verified at 400×700; desktop popover at 1280×720. Found + fixed a callout-scope bug. e2e scenarios added. |
| create-pr | ✓ | 1m 43s | Draft PR #642 + hickey/lowy analysis comment. |
| ci | ✓ | 1m 6s | Vira CI passed against HEAD e9d435bf. Built 5 packages incl. link-check (250 internal links). |
| **Total** | | **29m 27s** | |

**Slowest step**: `test` (7m 49s).

### Optimization suggestions

- **`test` dominated at 7m 49s** because chrome-mcp surfaced a real bug (callouts also needing `data-footnote-scope`), which pushed an extra edit-commit-reverify cycle into this step. That's the test step doing its job — no change recommended. The alternative (catching it only in e2e during `ci`) would have been longer since the fix would have landed during `ci` retry-mode.
- **`hickey+lowy` at 5m 26s** ran both sub-agents in parallel; individual agent wall-clock was ~3m each. If both had to re-run on a future touch, `--review-model=haiku` would trim this at the cost of shallower critique.
- **`police` at 4m 34s** is close-to-floor for a change of this size; most of the cost is the three passes (rules + fact-check + elegance). No slack to cut without dropping a pass.
- **`create-pr` at 1m 43s** includes a backtick-escape fix round-trip (`gh pr edit`). Worth remembering: single-quoted heredoc is load-bearing, but writing backticks *inside* the heredoc must stay bare — escaping them as `\`` produces literal backslashes.

Workflow completed at 2026-04-21 21:43 UTC.

docs/guide/markdown/embed.md
```diff
@@ -6,11 +6,12 @@ date: 2022-08-03
 
 # Embedding
 
-You can embed files, using `![[..]]` - a syntax inspired by [Obsidian](https://help.obsidian.md/Linking+notes+and+files/Embedding+files). The HTML can be fully customized for each embed types.
+You can embed files, using `![[..]]` - a syntax inspired by [Obsidian](https://help.obsidian.md/Linking+notes+and+files/Embedding+files). The HTML can be fully customized for each embed type[^parent].
 
 > [!warning] 
 > The embed wiki-link syntax must appear on a paragraph of its own, with no other text added next to it.[^blk] Recursive embeds are supported.
 
+[^parent]: A page-level footnote, rendered into the bottom list and also shown in the click-anywhere popup. The popup is a progressive enhancement — the bottom list stays usable when JS is off or on print.
```
srid: Remove this outdated footnote.

emanote/default/templates/components/footnote-popup.tpl
```diff
@@ -0,0 +1,266 @@
+<!-- Footnote popup. Footnote refs have no anchor href and no bottom list
+     renders on screen; the body content is emitted inside a hidden
+     <aside data-footnote-list>, sourced only by this script for the
+     popup clone. Bound to data-footnote-ref / data-footnote-id /
+     data-footnote-list / data-footnote-scope. -->
+
+<style data-category="footnote-popup">
```
srid, line 12: Fact check if we stricktly need all these styles in this file and that they cannot be done in tailwind itself.

docs/guide/markdown/embed.md
srid: Removed in 2a7c9238 — the copy described a progressive-enhancement fallback that no longer exists (the popup is the only on-screen UI now).

emanote/default/templates/components/footnote-popup.tpl
srid: Audited in 2a7c9238. Popover chrome, body styling, active-state, focus-visible, and the mobile bottom-sheet shape all moved to inline Tailwind utilities (applied via JS `className` for the popover/body, inline on the `<sup>` for state variants). The `.emanote-footnote-popup--mobile` class-toggle is gone — `max-sm:*` variants do it viewport-conditionally. The only CSS that survives is one rule: `#emanote-footnote-popover::backdrop { background: transparent; }`. Tailwind has no variant for the Popover `::backdrop` pseudo-element, and expressing it any other way would be strictly worse than one line of CSS.

PR #647, opened 2026-04-22
feat: add phase-1 MCP HTTP sidecar

### LINKED ISSUE #645: Add MCP server for live Emanote model
## Summary

Add an MCP server to Emanote that runs in the same long-lived process as the live server, so MCP clients can query Emanote's in-memory notebook model directly instead of going through the filesystem.

Uses the Haskell [`dpella/mcp`](https://github.com/dpella/mcp) library for protocol and transport support, while keeping Emanote's own model/query/export code as the source of truth.

## Goals

- Run live server and MCP server in the same process
- Reuse the existing reactive Emanote model
- Expose notebook data primarily as MCP resources
- Add a small set of read-only tools for structured queries
- Ship this incrementally, one PR per phase

## Non-goals

- No write/edit tools in the first iteration
- No separate notebook parser for MCP
- No filesystem-first MCP layer that bypasses Emanote's model
- No authentication today — the server is intended for local use bound to a loopback port

## Design direction

- Ema remains responsible for keeping the in-memory model live and for normal site rendering.
- MCP sits beside Ema, sharing the same current model snapshot.
- MCP does not go through `SiteRoute`/`Ema.Asset`; it reads from Emanote's model/query/export layer directly.
- Prefer MCP resources for notebook reads; use tools only for query-shaped operations.
- **Keep `resources/list` independent of notebook size.** Anything that scales linearly in N inflates context on every poll. Per-item addressing goes through RFC 6570 URI templates advertised on `resources/templates/list`.

### Phase 1: Same-process MCP transport + minimal handshake — ✅ Shipped in #648

Scope:

- Add `dpella/mcp` dependency
- Add CLI/config plumbing to enable MCP HTTP transport alongside live server (`emanote run --mcp-port PORT`)
- Start MCP endpoint in the same long-running process as the live server
- Implement minimum lifecycle/RPC surface (`initialize`, empty `resources/list`, empty `tools/list`, …)

### Phase 2: Notebook-backed resources — ✅ Shipped in #649

Scope:

- Wire MCP handlers to the live model via `Ema.Dynamic.currentValue` (returns a non-blocking `IO Model` reader; no shared `IORef`, no startup race)
- Expose notebook state as MCP resources:
  - `emanote://export/metadata` — JSON metadata for every note (titles, source paths, parent routes, resolved links); also the discovery surface for note paths
  - `emanote://note/{path}` — per-note read via an RFC 6570 URI template advertised on `resources/templates/list`
- Distinguish error codes: 400 for unrecognized URIs, 404 for catalog misses
- `resources/list` returns only the metadata export (O(1) in notebook size); per-note addressing goes through the template

Deliberately **not** shipped:

- Per-note entries in `resources/list` (would be O(N) per call)
- `emanote://export/content` (bundled-content blob — same information available via metadata + per-note reads, blows context budgets, the CLI `emanote export --format=content` still produces this artifact for human/script use)

> The `emanote://note/{path}` resource was later removed in phase 3 (#746). See the refinement note below.

### Phase 3: Query tools — ✅ Shipped in #745, refined in #746

Scope:

- Three read-only MCP tools so the model can navigate the notebook without dumping enumerations into context:
  - `find_notes` — case-insensitive substring search over note titles and source paths
  - `get_backlinks` — wrap `Model.Graph.modelLookupBacklinks`
  - `resolve_wikilink` — wrap the existing `Model.Link.Resolve` path
- Narrow input/output schemas; each tool result carries the note's notebook-relative `path` + `title`
- Reuse the same `IO Model` reader phase 2 plumbed

Refinement (#746):

- Drop the `emanote://note/{path}` per-note read resource introduced in phase 2 and the derived `uri` field on tool results. MCP clients (Claude Code, Codex, opencode, …) already have direct filesystem read tools, so serving note bodies over MCP just duplicates that capability and adds a round trip per read.
- The notebook-relative `path` in every tool result and in `emanote://export/metadata` (each note's `filePath`) is the canonical handle — clients hand it to their own `Read` tool. `resources/read emanote://note/{path}` now returns `400 Unrecognized resource URI`.
- Catalog dispatch collapses to a single `[(NotebookResource, Model -> ResourceBody)]` table; `listResources` and the new `readResource` lookup both derive from it, so adding a future resource is one tuple. Structural collapse cross-validated by [hickey + lowy](https://kolu.dev/blog/hickey-lowy/) lenses.

Acceptance:

- Clients can ask structured questions without pulling the metadata export blob
- Tool results are derived from the live model (same snapshot semantics as the metadata resource)
- A 422-note notebook is tractable through search instead of enumeration
- Wire surface is one resource + three tools; `path` is the canonical handle from MCP to the filesystem

### Phase 4: Live update support

Scope (smaller than originally outlined — phase 2's `currentValue` seam already eliminates the startup race the original plan was going to close):

- Wire MCP resource **subscriptions** (`resources/subscribe`, `resources/unsubscribe`, `notifications/resources/updated`)
- Emit `notifications/resources/list_changed` when the catalog changes (e.g. a new tag index resource appears)
- Cover at least the metadata export

Acceptance:

- A subscribed client observes updates after a notebook change without re-polling
- `capabilities.resources.subscribe` and `capabilities.resources.listChanged` flip from `Nothing` to enabled
- Polling clients are unaffected (subscription is opt-in)

### Phase 5: Optional prompt surfaces

Scope (only if the resource + tool surface proves insufficient):

- Add `prompts/list` / `prompts/get` for prompts like:
  - summarize note + backlinks
  - answer using notebook only
  - compare two notes

Acceptance:

- Prompts are optional sugar on top of the resource/tool API, not a replacement

## Settled open questions

- **CLI shape**: `emanote run --mcp-port PORT` (phase 1).
- **Model sharing**: `Ema.Dynamic.currentValue` returns `(IO Model, wrapped Dynamic)`; the reader is passed as a function argument to `MCP.run`, no shared `IORef` or new `EmanoteConfig` field (phase 2 commit `ff3b6407`).
- **Per-note bodies over MCP**: not served — clients read source files via their own filesystem tools, MCP hands them the relative `path` (phase 3, #746).
- **Authentication**: deferred — server is loopback-only, no auth today. Re-open if/when a remote-transport use case appears.

srid:
Closes #645

This lands phase 1 of Emanote MCP support using `dpella/mcp`. The scope is intentionally narrow: when the live server is running, `--mcp-port` starts an MCP HTTP endpoint in the same long-running process.

Phase 1 exposes only the MCP lifecycle plus a minimal read-only surface:
- `initialize`
- `resources/list`
- `resources/read`
- `tools/list`
- `tools/call`

It does not yet expose notebook-backed queries. That comes in later phases once we decide the Emanote-facing resource and tool model.

Verification:
- `nix develop -c cabal test test -j1`
- runtime smoke test against `POST /mcp` covering `initialize`, `resources/list`, `resources/read`, `tools/list`, and `tools/call`


srid:

| 1 | Hickey | Remove warning noise from the new MCP phase-1 module | Fixed in this PR |

### Hickey rationale
The phase-1 MCP addition stays structurally small: one new module owns the protocol surface, and the runtime integration is a narrow seam in `Emanote.run`. The only issue found in the review pass was avoidable warning noise in the new module, which would make the boundary feel rougher than it is. That was cleaned up in follow-up commit `07522021`.

### Lowy rationale
No boundary-level blocker showed up in phase 1. The volatile concern here is the MCP-facing transport and capability surface, and that is already isolated into `Emanote.MCP` instead of being smeared through the Emanote model/runtime code. Notebook-backed querying is still deferred, which is the right cut for this phase.

srid:

| sync | ✓ | 0s | git fetch ok; forge=github; noGit=false  |
| research | ✓ | 2m 1s | Issue #645 phase 1 confirmed; runtime seam is Emanote.run/emanoteSiteInput; dpella/mcp v0.3.0.0 provides HTTP handlers for initialize/resources/tools; implementation will add an optional --mcp-port live-server sidecar in the same process with a minimal read-only resource/tool surface.  |
| branch | ✓ | 9s | Created feature branch feat/mcp-phase-1 from master.  |
| implement | ✓ | 23s | Added phase-1 MCP HTTP sidecar wiring, minimal resource/tool handlers, dependency wiring in Cabal/Nix, and handler tests.  |
| check | ✓ | 5m 33s | nix develop -c cabal build emanote -j1 succeeded after wiring dpella/mcp through the flake package set and fixing MCP integration compile errors.  |
| docs | ✓ | 16s | Updated docs/start.md with the new --mcp-port live-server sidecar flag and clarified that notebook-backed MCP resources arrive in a later phase.  |
| fmt | ✓ | 19s | just fmt passed after one rewrite pass from cabal-fmt and fourmolu.  |
| commit | ✓ | 0s | Committed feat: add phase-1 MCP HTTP sidecar as 86d3e32e and pushed feat/mcp-phase-1 to origin.  |
| hickey+lowy | ✓ | 34s | Local hickey/lowy pass found no boundary-level blockers; applied one small cleanup commit (07522021) to remove warning noise from the new MCP module and pushed it.  |
| police | ✓ | 14s | Manual code-police sweep found no additional correctness or simplicity issues beyond the warning-noise cleanup already addressed in 07522021.  |
| test | ✓ | 2m 43s | nix develop -c cabal test test -j1 passed; live server + MCP sidecar smoke-tested on POST /mcp with initialize, resources/list, resources/read, tools/list, and tools/call succeeding.  |
| create-pr | ✓ | 10s | Draft PR #647 already existed; verified title/body scope and posted the required Hickey/Lowy analysis comment at https://github.com/srid/emanote/pull/647#issuecomment-4300089201.  |
| ci | ✓ | 20s | nix run github:juspay/vira ci -- -b passed for HEAD 00bd0b19afc994a40c4a2f4070bdb1130070537f.  |
| **Total** | | **25m 49s** | |

### Optimization suggestions

- The slowest step was `check` at 5m 33s. If phase 2 touches only the MCP seam again, it is worth adding or documenting a narrower Haskell build target than the current full `cabal build emanote -j1` path.
- `test` took 2m 43s because it combined unit coverage with a live MCP smoke test. If this pattern repeats, scripting the JSON-RPC smoke test into a reusable check would make later phases cheaper and less manual.
- `research` took 2m 1s mostly to pin down the `dpella/mcp` surface and seam into Emanote. Future MCP phases should reuse the current phase split and module boundary instead of re-evaluating transport/library choices.

Workflow completed at 2026-04-22T17:31:20-04:00.

srid:
## [Hickey/Lowy](https://kolu.dev/blog/hickey-lowy/) follow-up review

| 1 | Hickey | Hard-coded MCP version duplicates the canonical package version in [Emanote.MCP](./emanote/src/Emanote/MCP.hs) | Fix in this PR |
| 2 | Hickey | Phase-1 MCP registry is fragmented across lists, dispatch, and self-report payload in [Emanote.MCP](./emanote/src/Emanote/MCP.hs) | Fix in this PR |
| 3 | Lowy | `Emanote.MCP` mixes HTTP/bootstrap volatility with phase-surface volatility | Fix in this PR |

### Hickey rationale
Two structural issues stood out.

First, the MCP server version is hard-coded as a string literal in `Emanote.MCP`, even though Emanote already has a canonical package-derived version source. That creates two authorities for one fact and invites drift on later releases.

Second, the phase-1 MCP surface is represented by convention across separate metadata lists, dispatch functions, handlers, and payload construction. The invariant is effectively “everything listed must also be handled,” but the code does not encode that invariant directly. That raises the reasoning load for every later MCP addition.

### Lowy rationale
The new `Emanote.MCP` module currently owns two independent reasons to change:

- HTTP/bootstrap/runtime hosting details
- phase-surface semantics for the exposed MCP resources and tools

That boundary is likely to get hit from both sides in later phases. Expanding notebook-backed MCP behavior should not need to edit the same module that owns HTTP server startup. A cleaner split would keep transport/bootstrap separate from the phase-specific resource/tool contract.

### Suggested follow-up
- Source the advertised MCP implementation version from `Paths_emanote.version`
- Collapse the phase-1 resource/tool definitions into a single registry representation and derive listing/dispatch from it
- Split transport/bootstrap from phase-surface code, e.g. `Emanote.MCP.HTTP` and `Emanote.MCP.Phase1`

emanote/src/Emanote/MCP.hs
```diff
@@ -0,0 +1,148 @@
...
+import MCP.Server
+import MCP.Server.HTTP
+import MCP.Types qualified as MT
+import Relude
+
+phase1ResourceUri :: Text
+phase1ResourceUri = "emanote://phase-1/info"
+
+phase1ToolName :: Text
+phase1ToolName = "emanote_phase_1_status"
```
srid: Instead of these dummy stuff, we should return real information ... something trivial.

emanote/src/Emanote.hs
```diff
@@ -72,8 +74,18 @@ run cfg@EmanoteConfig {..} = do
   case CLI.cmd _emanoteConfigCli of
     CLI.Cmd_Ema emaCli -> do
       let emaCfg = SiteConfig emaCli def
-      Ema.runSiteWith @SiteRoute emaCfg cfg
-        >>= postRun cfg
+      if Ema.CLI.isLiveServer (Ema.CLI.action emaCli)
```
srid, line 77: Add a `-- HACK` comment here because ideally we want to support `emanote run --mcp-port`

emanote/src/Emanote/MCP.hs
srid: Addressed on the branch.

The dummy phase text is gone. The MCP surface now returns real server information instead:
- resource: `emanote://server/info`
- tool: `emanote_get_server_info`
- payload/text include server `name`, `title`, package-derived `version`, `protocolVersion`, plus the exposed resource/tool names

Relevant code:
- `Emanote.MCP.Info`
- `Emanote.MCP.Surface`
- commit `afb04974` split the server/surface and switched to real info
- commit `fa0d6a42` adds the module-level comments

PR #650, opened 2026-04-24
Atom feed: empty query results no longer crash the build

### LINKED ISSUE #490: RSS feed generation crashes on empty query results
```
emanote: Bad feed: LMLRoute_Md R[/en/blog.md]: can't find note query
```

https://github.com/nixos-asia/website/actions/runs/7403841677/job/20144391105

srid:
**A `feed.enable: true` note whose `query` matches no notes — say a blog with no posts yet — used to abort the build** with `emanote: Bad feed: ... no notes matched the query`. `Emanote.View.Template.renderResourceRoute` was calling `error` whenever `renderFeed` returned a `Left`, treating a runtime emptiness state as a fatal program error.

`renderFeed` keeps its `Either LText LByteString` signature so configuration mistakes (missing/invalid `query` block, multiple queries, missing `page.siteUrl`) still surface as `Left` and **still fail the build loud** — those are user errors that need to be visible. Only the `runQuery [] -> empty-feed` case is rerouted: when a query is syntactically valid and present but the model has no matching notes, `renderFeed` now returns a minimal valid Atom document (id derived from the note's route, title from the note's title, date from `_noteMeta.date` or a 1970 fallback) via a new `renderEmptyFeed` helper. _The fix narrowed scope after the first review pass — earlier commits on this branch over-corrected by absorbing every `Left` and have since been reverted at the policy level._

Test coverage is in `Emanote.View.FeedSpec` (5 cases pinning down `getNoteQuery`'s failure classifications + 1 case for `renderEmptyFeed`'s XML output), plus a Cucumber e2e scenario that fetches `/empty-feed.xml` from a fixture whose query matches nothing — exercised in both live and static modes.

Closes #490.

### DIFF
--- a/emanote/CHANGELOG.md
+++ b/emanote/CHANGELOG.md
@@ -16,6 +16,10 @@
 - Mermaid: add `elk` layout ([#618](https://github.com/srid/emanote/pull/618))
 - Home Manager module: macOS support via launchd ([#623](https://github.com/srid/emanote/pull/623))
 
+**Fixes**
+
+- Atom feed: misconfigured feeds (missing query block, empty query results) no longer crash the whole build; an empty-but-valid Atom document is emitted instead ([#490](https://github.com/srid/emanote/issues/490))
+
 ## 1.4.0.0 (2025-08-18)
 
 **Notable features**
--- a/emanote/CHANGELOG.md
+++ b/emanote/CHANGELOG.md
@@ -19,7 +19,7 @@
 
 **Fixes**
 
-- Atom feed: misconfigured feeds (missing query block, empty query results) no longer crash the whole build; an empty-but-valid Atom document is emitted instead ([#490](https://github.com/srid/emanote/issues/490), [#650](https://github.com/srid/emanote/pull/650))
+- Atom feed: a feed query that matches no notes no longer crashes the build; an empty-but-valid Atom document is emitted instead. Configuration errors (missing/invalid query block, missing `page.siteUrl`) still fail loudly ([#490](https://github.com/srid/emanote/issues/490), [#650](https://github.com/srid/emanote/pull/650))
 
 ## 1.4.0.0 (2025-08-18)
 
--- a/emanote/CHANGELOG.md
+++ b/emanote/CHANGELOG.md
@@ -18,7 +18,7 @@
 
 **Fixes**
 
-- Atom feed: misconfigured feeds (missing query block, empty query results) no longer crash the whole build; an empty-but-valid Atom document is emitted instead ([#490](https://github.com/srid/emanote/issues/490))
+- Atom feed: misconfigured feeds (missing query block, empty query results) no longer crash the whole build; an empty-but-valid Atom document is emitted instead ([#490](https://github.com/srid/emanote/issues/490), [#650](https://github.com/srid/emanote/pull/650))
 
 ## 1.4.0.0 (2025-08-18)
 
--- a/emanote/emanote.cabal
+++ b/emanote/emanote.cabal
@@ -255,3 +255,4 @@ test-suite test
       Emanote.Pandoc.ExternalLinkSpec
       Emanote.Pandoc.Renderer.CalloutSpec
       Emanote.Route.RSpec
+      Emanote.View.FeedSpec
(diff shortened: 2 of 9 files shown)


srid:

| # | Lens   | Finding                                                  | Disposition       |
|---|--------|----------------------------------------------------------|-------------------|
| 1 | Hickey | Module-split error handling (Template/Feed)              | Fixed in this PR  |
| 2 | Hickey | Cascading `maybe mempty` in fallback (silent zero bytes) | Fixed in this PR  |
| 3 | Hickey | No observability on swallowed errors                     | Deferred          |
| 4 | Lowy   | Fallback policy collapses unrelated failure modes        | Deferred          |
| 5 | Lowy   | FeedSpec doesn't exercise `renderFeed` under failure     | Deferred          |
| 6 | Lowy   | Affirmed: interface stable under feed-volatility changes | No-op             |
| 7 | Lowy   | Affirmed: single call site is not a decomposition signal | No-op             |

### Hickey rationale

The elimination of the module-split error handling is the right structural move. The old code braided two distinct decisions across two modules: Feed.hs signaled failure via `Left`, Template.hs decided what to do via `error`. Template.hs had no business knowing what constitutes a bad feed. That braiding is now gone.

The `maybe mempty encodeUtf8` in the original fallback would have silently emitted zero bytes if `Export.textFeed` returned `Nothing` — a second swallow nested inside the first. Replaced with `fromMaybe (error "...")` since inputs to `Atom.nullFeed` are fully hardcoded; a `Nothing` here is a library bug, not a user error, and should fail loud.

Observability of the suppressed `Left _err` was raised but is deferred: `renderResourceRoute` is pure, relude deprecates `trace` in committed code, and proper logging needs threading a `MonadLogger` through. Out of scope for a crash fix.

### Lowy rationale

`renderFeed` now encapsulates feed-generation volatility correctly. Before, error policy crossed the module boundary into Template.hs; after, all failure paths and the fallback strategy are internal to Feed.hs. Template.hs will not change when feed-generation volatility changes — adding new failure modes (size limits, etc.) or swapping the fallback strategy (404 instead of empty feed) leaves it untouched.

`renderEmptyFeed` belongs in Feed.hs — it is part of the fallback _strategy_ for feed rendering, not a generic utility.

The collapse of five distinct failure paths (missing query, invalid query, no matches, missing `page.siteUrl`, library serialization) into a single silent fallback is real and flagged — but actioning it requires the same logger threading deferred under finding 3. Same followup.

`renderFeed` under failure is not exercised end-to-end in `FeedSpec` because constructing a `Model` in unit tests is non-trivial. The pure `getNoteQuery` and `renderEmptyFeed` tests cover the classification and the bytes; the full integration is deferred.

srid:

| research | ✓ | 3m 49s | Identified crash at Template.hs:99 |
| branch | ✓ | 15s | On feature branch fix/feed-crash-490 |
| implement | ✓ | 2m 27s | Added FeedSpec; renderFeed falls back; Template.hs no longer errors |
| check | ✓ | 1m 7s | cabal build all succeeded |
| docs | ✓ | 22s | CHANGELOG entry under 1.6.0.0 Fixes |
| fmt | ✓ | 26s | fourmolu, cabal-fmt, hlint, nixpkgs-fmt all passed |
| commit | ✓ | 40s | Primary feature commit 33134ef8 |
| **hickey+lowy** | ✓ | **5m 21s** | Bulletproof fallback applied (95b5ec74); observability deferred |
| police | ✓ | 2m 53s | Rules: 1 deferred. Fact-check: clean. Elegance: 2 fixes (4bdf376b, fe4ea0cb) |
| test | ✓ | 10s | 29 examples, 0 failures |
| create-pr | ✓ | 48s | Draft PR + hickey/lowy comment |
| ci | ✓ | 4m 16s | All GitHub Actions green on HEAD |
| **Total** | | **25m 0s** | |

**Slowest step**: `hickey+lowy` (5m 21s)

### Optimization suggestions

- **`hickey+lowy` dominates at 5m 21s.** The two structural-review sub-agents now run in parallel via the `Agent` tool (not the `Skill` tool, which would serialize). Further savings would come from passing `--review-model=haiku` for tasks of this size — sonnet/opus give little extra value on a five-line crash fix.
- **`research` at 3m 49s** ate context exploring `Feed.hs`, `Template.hs`, `Note.hs`, the test infrastructure, and the model construction surface. For a future crash-bug fix, pre-citing the specific `error` call site in the prompt would skip most of that mapping.
- **`ci` at 4m 16s** is dominated by the `e2e-tests` matrix (live + static, ~4m each). Consider a `--from ci-only` re-run path if iterating on CI-only changes.
- **`police` at 2m 53s** ran the three-pass elegance loop and turned up two real fixes (date-helper reuse, Haddock trim). Worth keeping; the cost was paid for genuine improvements to the diff.

Workflow completed at 2026-04-24T23:55Z.

PR #651, opened 2026-04-24
Resolve relative links from index.md against its own folder

### LINKED ISSUE #608: Relative paths in an folder/index.md file
**Describe the bug**

In an index.md file in a subfolder, normal links relative to the file do not work as if the index.md is in that subfolder, but rather that the index.md is in the parent dir. 

**To Reproduce**
1. Create an index.md in a subfolder
2. Add a relative link in that subfolder
3. See the broken link
4. Adjust the relative link as if index.md is in the parent folder
5. See the working link

See my [fork of the emanote template](https://github.com/IllustratedMan-code/emanote-template) for an example

**Expected behavior**
Relative links should work as if the index.md is where it is.

**Comments**
This is somewhat addressed in the docs:

> Instead of foo.md, you can also use foo/index.md (or index.org if using [Org Mode ✍️](https://emanote.srid.ca/guide/orgmode)) or foo/index.yaml. Internally, Emanote will treat them as foo.md and foo.yaml respectively.

I would argue that this is a bit confusing though. I'm happy to make a pull request with a more detailed explanation, but how hard would it be to make `index.md` use the correct relative path? This way editor integration would be better (i.e. path suggestions).

srid:
**Relative URLs inside `<dir>/index.md` now resolve against `<dir>/`, not its parent.** Closes #608. Previously, Emanote canonicalized `subfolder/index.md` to route `subfolder` (dropping the trailing `index` slug), and then derived the link-resolution base by taking that route's parent — which landed one folder too high. So `[foo](./foo.md)` inside `subfolder/index.md` resolved to `foo.md` at the root instead of `subfolder/foo.md`.

The fix derives the base from the note's on-disk source filepath (`takeDirectory`) rather than its canonicalized route. A new `noteResolveLinkBase :: Note -> Maybe (R 'Folder)` lives in `Emanote.Model.Note`, and `modelResolveLinkBase :: Model -> LMLRoute -> Maybe (R 'Folder)` in `Emanote.Model.Type` wraps the model lookup so the four splice callsites (`Url.urlResolvingSplice` and three in `Pandoc.Renderer.Embed`) collapse to a single function call. *Synthesized notes (placeholders, missing-note pages) lack an on-disk path and fall back to `noteParent` — they don't trigger the bug since they have no canonicalized source to drift from.*

Five new unit tests in `RelSpec` cover top-level files, subfolder index files, deeply nested index files, sibling non-index files, and an end-to-end `parseUnresolvedRelTarget` resolution from `subfolder/index.md`. An additional **e2e scenario** runs in both `live` and `static` modes against a `subfolder/index.md` + `subfolder/sibling.md` fixture, asserting the article-body anchor's href resolves to `subfolder/sibling.html`. *The selector is scoped to `<article>` because the sidebar nav surfaces a child link to the same target — and that path is auto-generated from the route hierarchy, so it stays correct even when the bug is live.*

> The post-implement Hickey review flagged the lookup-and-fallback algorithm as inlined identically at four splice sites. The second commit extracts `modelResolveLinkBase` to encode the rule once. Lowy review was clean.

### DIFF
--- a/emanote/CHANGELOG.md
+++ b/emanote/CHANGELOG.md
@@ -16,6 +16,10 @@
 - Mermaid: add `elk` layout ([#618](https://github.com/srid/emanote/pull/618))
 - Home Manager module: macOS support via launchd ([#623](https://github.com/srid/emanote/pull/623))
 
+**Bug fixes**
+
+- Resolve relative URLs inside `<dir>/index.md` against `<dir>/` instead of its parent ([#651](https://github.com/srid/emanote/pull/651), closes [#608](https://github.com/srid/emanote/issues/608))
+
 ## 1.4.0.0 (2025-08-18)
 
 **Notable features**
--- a/emanote/src/Emanote/Model/Link/Rel.hs
+++ b/emanote/src/Emanote/Model/Link/Rel.hs
@@ -10,7 +10,7 @@ import Data.IxSet.Typed qualified as Ix
 import Data.List.NonEmpty qualified as NEL
 import Data.Map.Strict qualified as Map
 import Data.Text qualified as T
-import Emanote.Model.Note (Note, noteDoc, noteRoute)
+import Emanote.Model.Note (Note, noteDoc, noteResolveLinkBase, noteRoute)
 import Emanote.Route (LMLRoute, ModelRoute)
 import Emanote.Route qualified as R
 import Emanote.Route.SiteRoute.Type qualified as SR
@@ -71,7 +71,7 @@ noteRels note =
         $ flip concatMap (Map.toList m)
         $ \(url, instances) -> do
           flip mapMaybe (toList instances) $ \(attrs, ctx) -> do
-            let parentR = R.withLmlRoute R.routeParent $ note ^. noteRoute
+            let parentR = noteResolveLinkBase note
             (target, _manchor) <- parseUnresolvedRelTarget parentR attrs url
             pure $ Rel (note ^. noteRoute) target ctx
 
--- a/emanote/src/Emanote/Model/Note.hs
+++ b/emanote/src/Emanote/Model/Note.hs
@@ -32,7 +32,7 @@ import Network.URI.Slug (Slug)
 import Optics.Core ((%), (.~))
 import Optics.TH (makeLenses)
 import Relude
-import System.FilePath (takeFileName, (</>))
+import System.FilePath (takeDirectory, takeFileName, (</>))
 import Text.Pandoc (readerExtensions, runPure)
 import Text.Pandoc.Builder qualified as B
 import Text.Pandoc.Definition (Pandoc (..))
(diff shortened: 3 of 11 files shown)


srid:

| 1 | Hickey | Four-site duplication of "lookup note; if found use noteResolveLinkBase, else fall back to routeParent" | Fixed in this PR |
| 2 | Lowy | (no findings) | — |

### Hickey rationale

The unenforced rule "when resolving links for a note identified by `noteRoute`, look up the note; if found, use `noteResolveLinkBase`; otherwise fall back to `R.withLmlRoute R.routeParent noteRoute`" was encoded identically at four locations: three splices in `Pandoc.Renderer.Embed` (`embedBlockWikiLinkResolvingSplice`, `embedBlockRegularLinkResolvingSplice`, `embedInlineWikiLinkResolvingSplice`) and one in `Pandoc.Renderer.Url` (`urlResolvingSplice`). The rule lived only as a social contract — a future change to the algorithm would require updating four sites, and missing one would produce a silent divergence. Extracted to `modelResolveLinkBase :: ModelT f -> LMLRoute -> Maybe (R 'Folder)` in `Emanote.Model.Type` so each callsite reduces to a single function call. (Commit `af887c47`.)

### Lowy rationale

Volatility is correctly encapsulated. `noteResolveLinkBase` isolates the volatile decision — "derive link-base from on-disk path when available, fall back to route hierarchy otherwise" — behind a stable interface. That decision is an observed axis of change (the bug itself), not speculative. Placement in `Note.hs` is correct: the function is a pure computation over `Note` attributes, and putting it in `Rel.hs` would conflate link-base derivation (an input property) with link resolution. The fallback for synthesized notes (no `_noteSource`) is sound — synthesized notes are not the class of notes the bug affected, and `noteParent` is the only signal available. Reuse across four independent splice callsites is the pattern expected of a volatility boundary.

srid:

| research | ✓ | 5m 28s | Located bug at Rel.hs:74 + Url.hs:32 + Embed.hs:34/51/62. Fix: derive parent from `_noteSource` filepath instead of canonicalized route. |
| branch | ✓ | 4s | On branch fix-index-relative-links from origin/master. |
| implement | ✓ | 7m 36s | Added noteResolveLinkBase + resolveLinkBaseFromFilePath in Note.hs; updated 4 callsites; 4 unit tests in RelSpec.hs all pass. |
| check | ✓ | 5s | cabal build all: up to date, no errors. |
| docs | ✓ | 12s | Existing docs accurate; no update needed. |
| fmt | ✓ | 25s | pre-commit: cabal-fmt, fourmolu, hlint, nixpkgs-fmt all pass. |
| commit | ✓ | 21s | Commit feec2723 pushed. |
| hickey+lowy | ✓ | 6m 23s | Hickey: 1 finding (4-site dedup) → fixed in af887c47. Lowy: no findings. |
| police | ✓ | 3m 59s | Rules clean (7/7); fact-check clean; 1 deferred efficiency concern. |
| test | ✓ | 10s | 28/28 tests pass. |
| create-pr | ✓ | 41s | Draft PR #651 with hickey/lowy comment. |
| ci | ✓ | 4m 27s | All 3 CI checks green on HEAD af887c47. |
| **Total** | | **31m 47s** | |

**Slowest step**: `implement` (7m 36s)

### Optimization suggestions

- **`implement` dominated**: research and implement together took 13 minutes. Splitting the bug fix from the test scaffolding into separate iterations (write the failing test, then write the smallest possible fix) might shave time, but for a bug-fix-with-helper-extraction this is roughly the floor.
- **`hickey+lowy` (6m 23s)**: ran sequentially rather than in parallel. Future runs should emit both `Agent` tool calls in a single response — easily halves this step.
- **`ci` (4m 27s)**: includes Playwright browser cache hit + e2e tests. For followup pushes on this branch, `--from ci-only` would skip everything before CI.
- **`police` (3m 59s)**: three lenses ran serially against the diff. `/simplify` already runs its three lenses in parallel; the rules+fact-check passes are bounded by main-context work. Acceptable.

PR #652, opened 2026-04-25
Nested + foldable Obsidian callouts

srid:
**The remaining two items in #465 are now in.** `> [!type]+` and `> [!type]-` give you foldable callouts (initially expanded and collapsed, respectively); nested callouts work by indenting an inner blockquote with `> >`. Closes the last two unchecked boxes on the issue.

Foldable callouts render as native `<details>`/`<summary>` so the toggle works without JavaScript, and the chevron rotates via a CSS `[open]` selector. Nested callouts mostly came for free — the recursive `pandocSplice` already re-applies the block renderer to inner blocks, so the parser only needed to keep the inner `BlockQuote` intact in `Callout.body`. The visible "feature" there is really the docs section plus a CSS rule that drops the trailing margin on the last inner callout so it hugs the parent body.

The header parser is split into `parseCalloutHeader :: Text -> Maybe (CalloutType, Maybe FoldState)` returning both pieces; `parseCalloutType` is preserved as a back-compat shim for the existing tests. _A defensive `absorbFoldSuffix` merges adjacent `Str "+"`/`Str "-"` tokens onto the header in case Pandoc tokenises `[!tip]+` as two `Str` nodes; the comment spells out why we deliberately do **not** absorb across `Space`._

Reviewed by `/hickey` and `/lowy`. One comment-clarification fix landed (commit `007c722c`) and one CSS dead-selector fix from `/code-police`'s fact-check pass (`e3497610`). Two structural items deferred — see the [Hickey/Lowy](https://kolu.dev/blog/hickey-lowy/) Analysis PR comment for the disposition table.

> **Deferred:** the three-branch template duplicates the `<details>` wrapper between expanded/collapsed (Heist doesn't expose conditional attribute presence cleanly without losing access to template-level `color`/`icon` binds), and the splice/template split for fold state isn't compiler-enforced for exhaustiveness. Both are quality-of-life refactors — neither is structurally wrong.

### Try it locally

```sh
nix run github:srid/emanote/dizzy-king -- -L docs run
```

Then visit `/callout` to see the new `Foldable callouts` and `Nested callouts` sections render.


srid:

| 1 | Hickey | Unvalidated callout type can crash via `lookupHtmlTemplateMust` | No-op — predates this PR; an allowlist would break the documented custom-callout extension point (#573) |
| 2 | Hickey | `absorbFoldSuffix` doesn't handle `Space, Str "-"` form | No-op — absorbing across `Space` would mis-parse `> [!tip] -` (a tip with a hyphen-prefixed title) as foldable; comment now spells this out |
| 3 | Hickey | Three template branches duplicate ~90% of the wrapper structure | Deferred — Heist doesn't support conditional attribute presence in `${...}` interpolation, and a Haskell-side wrapper splice loses access to template-level `color`/`icon` binds |
| 4 | Hickey | `FoldState` exhaustiveness isn't enforced across Haskell+template | Deferred — same shape as #3; would require consolidating the three-way dispatch into a single Haskell case |
| 5 | Lowy | `absorbFoldSuffix` encoded an undocumented Pandoc tokenisation assumption | Fixed in this PR (007c722c) |
| 6 | Lowy | `Maybe FoldState` slightly indirect vs. flat ADT | No-op — correct in context; flattening is a stylistic preference, not a boundary issue |
| 7 | Lowy | `callout:fold-state` data attribute as a customiser hook | No-op — Lowy endorsed this as the right seam; `data-callout-fold` is a stable string handle that degrades gracefully if a new state is added |
| 8 | Lowy | Nested-callout CSS selector tightly coupled to DOM shape | Fixed in this PR (e3497610) — dead second selector removed; the surviving one matches both foldable and non-foldable parents |
| 9 | Lowy | Custom callout types extension point unaffected by changes | No-op — type templates just `<apply template="_base">`, so the new conditionals inherit automatically |

### Hickey rationale

The three-branch template — `<div>` for non-foldable, `<details open>` for foldable+expanded, `<details>` for foldable+collapsed — is structurally justified: these are three distinct HTML elements with different semantics. The duplication cost is the price of branching at the template layer rather than at the Haskell layer; the alternative (always-`<details>` plus CSS to disable the toggle) would complect the DOM with a presentational workaround and obscure the accessibility intent. The shared content (`callout-icon-svg`, `callout-fold-marker`, `callout-body-block`) is already factored via Heist `<bind>`. Pushing dedup further is a quality-of-life cleanup rather than a structural correction.

### Lowy rationale

Three volatility axes are encapsulated independently: parsing (`parseCalloutHeader` + `absorbFoldSuffix`), render orchestration (the three conditional Heist splices), and presentation (CSS in `styles.tpl`). Each can change without ripple — if Obsidian extends the fold-suffix grammar, only the parser changes; if a new fold state is added, only the orchestration plus a CSS rule. The custom-callout extension point (#573) is unaffected because individual type templates delegate to `_base.tpl` without knowing about fold state. The `callout:fold-state` data attribute is the right customiser seam — a stable string handle that downstream CSS/JS can hook into without touching Haskell internals.

srid:

| research | ✓ | 6m 10s | Mapped callout implementation; identified 4 files to change |
| branch | ✓ | 16s | On feature branch dizzy-king |
| implement | ✓ | 9m 20s | Parser + Callout.foldState + 3-branch template + CSS; nested verified end-to-end via `emanote gen` |
| check | ✓ | 5s | cabal build all clean |
| docs | ✓ | 27s | callout.md + CHANGELOG.md updated |
| fmt | ✓ | 30s | cabal-fmt, fourmolu, hlint, nixpkgs-fmt all clean |
| commit | ✓ | 31s | 887a62b6 (6 files, +219/-26) pushed |
| hickey+lowy | ✓ | 6m 15s | Lowy 1 fixed (007c722c); Hickey 3+4 deferred; rest no-op |
| police | ✓ | **17m 31s** | Fact-check: dead CSS selector fixed (e3497610). Elegance: `whenSplice` → `Heist.Splices.ifISplice` (0031e5b0) |
| test | ✓ | 30s | cabal test: 32 examples 0 failures (incl. 9 new callout parser tests) |
| create-pr | ✓ | 1m 50s | Draft PR #652 + Hickey/Lowy disposition table comment |
| ci | ✓ | 4m 5s | e2e (live) 4m20s, e2e (static) 4m11s, flake-parts-docs 1m49s — all green on HEAD `0031e5b0` |
| done | ✓ | 0s | Workflow complete |
| **Total** | | **48m 3s** | |

### Optimization suggestions

- **`police` dominates at 36% of total** (17m 31s). The bulk was the `/simplify` Phase 2 spawning three reviewer agents in parallel, then code-search inside each — for a diff this small, a single inline pass would have been faster than the multi-agent fan-out.
- **`research` (6m 10s) and `hickey+lowy` (6m 15s)** ran similar lengths. Combined they're 26%. Pre-reading `Emanote/Pandoc/Renderer/Callout.hs` and the `_base.tpl` callout template before invoking `/do` would shave the research step substantially.
- **`implement` (9m 20s)** included one fmt-induced rebuild and an end-to-end render verification via `emanote gen`. The render check was worth the time (caught zero issues here, but would have caught template-XML errors immediately).
- **CI (4m 5s)** was the unavoidable e2e wait. For follow-up commits to this PR, `--from ci-only` skips everything else and just re-runs CI against the new HEAD.

Workflow completed at 2026-04-25T00:25:33Z.

Issue #655, opened 2026-04-25
Pandoc renderers: support IO at render time


srid:
Follow-up to #625 (inline mermaid SVG rendering).

The custom Pandoc renderer interface (`PandocBlockRenderer`/`PandocInlineRenderer`) returns `Maybe (HI.Splice Identity)` — pure splices that cannot perform IO. This forced the mermaid renderer in #625 to live at parse time instead of render time, with consequences:

- **Parse-time IO**: subprocess + tempdir + filesystem during note parsing. Parsing is no longer a pure transformation.
- **Cache invalidation coupling**: the cached AST contains rendered output. A theme/CSS-only change can't reuse it; an `mmdc` upgrade doesn't invalidate the AST.
- **Test friction**: `parseNote` now requires `MonadIO` plus a real `mmdc` binary or careful mocking.

Surfaced by Hickey-lens review during /do on #625 — ease (existing pure interface is at-hand) winning over simplicity (rendering belongs at render time).

## Proposal

Widen the renderer interface to permit IO, e.g. `Maybe (HI.Splice IO)` or a custom monad with `MonadIO` + cache access. Migrate the mermaid transform from parse-time to render-time. With #654 (content-addressed cache) in place, the render-time path is no slower than parse-time on incremental builds — and decouples diagram rendering from note re-parsing.

This is a wider refactor than mermaid alone; other splices that currently hold IO at arm's length (Stork query, embed) may benefit too.

Issue #654, opened 2026-04-25
Mermaid: content-addressed cache for rendered SVGs


srid:
Follow-up to #625 (inline SVG rendering).

Every build re-invokes `mmdc` per diagram. `mmdc` startup is 1-3s (Node + Puppeteer + headless Chromium), so a note with N diagrams adds N × 2s. Live-server mode reparses the entire note on every save: editing prose around a single diagram still re-renders all diagrams in that note.

Surfaced by structural review during /do on #625 (Hickey — render and build are two folds; Lowy — cache invalidation is its own volatility axis; efficiency review — the actual hot-path problem).

## Proposal

Content-addressed cache keyed on SHA(mermaid source ++ mmdc version):

```haskell
type MermaidCache = TVar (HashMap (Digest SHA256) Text)
```

- Constructed once alongside `IndexVar`.
- Threaded through `parseNote` → `transformMermaidBlocks` like `scriptingEngine`.
- In `runMmdc`: hash code, lookup, return cached SVG; on miss, render and insert.
- No eviction needed — keys are content hashes, total size bounded by unique diagrams in corpus.

Optional: persist to disk under the build cache dir for cross-build reuse (CI-friendly).

Issue #653, opened 2026-04-25
Mermaid: configurable rendering mode (server/client/off)


srid:
Follow-up to #625 (inline SVG rendering).

The current implementation is "always-on if `mmdc` is on PATH". This makes per-note behaviour depend on the host environment — same source produces different output on a dev laptop with `mmdc` installed vs. CI without it. The retained `js.mermaid` snippet for client-side rendering is silently dead when `mmdc` is also present, since the code blocks no longer exist for the JS to find.

Surfaced by structural review during /do on #625 (Hickey lens — environment complected with semantics; Lowy lens — sequence-volatility unencapsulated).

## Proposal

Add an explicit knob, e.g.:

```yaml
template:
  mermaid:
    renderer: server  # server | client | off
```

- `server` requires `mmdc` and errors loudly if absent (no silent fallback).
- `client` skips the AST transform; the existing `js.mermaid` snippet handles rendering.
- `off` leaves code blocks intact with no rendering.

Default: `server` when `mmdc` is detected at startup, otherwise `client`. The default itself remains environment-derived, but the user can pin it explicitly per site or per note.

PR #656, opened 2026-04-25
Render mermaid diagrams to inline SVG at build time

### LINKED ISSUE #119: <div></div> cannot used in raw HTML
**Describe the bug**
Placing a `<div></div>` inside a block using ```{=html} in a emanote markdown file causes the following error: 

```
div cannot contain text looking like its end tag
CallStack (from HasCallStack):
  error, called at src/Text/XmlHtml/HTML/Render.hs:132:9 in xmlhtml-0.2.5.2-CTNpBEzbTM9I9R1TWdgm5i:Text.XmlHtml.HTML.Render
```

**To Reproduce**
1. Create the following markdown file
```
# Demo

Hello World

```{=html}
<div>
Hello World
</div>
\```
```
(Without the slash, GitHub's markdown makes the slash necessary).

2. Run `emanote`

**Expected behavior**
The page renders two Hello Worlds.

 - Browser Firefox
 - Version 91

**Additional context**
The following code in xmlhtml seems to be relevant: https://github.com/snapframework/xmlhtml/blob/master/src/Text/XmlHtml/HTML/Render.hs#L133

This came up because I was trying to follow [these instruction for embedding an Argdown component](https://argdown.org/guide/embedding-your-maps-in-a-webpage.html) which include `<div>` tags.

### LINKED ISSUE #625: Render Mermaid diagrams to inline SVG at HTML render time
## User-facing feature

Emanote should support offline Mermaid diagrams by rendering Mermaid code blocks to inline SVG during HTML generation.

Authors should be able to write ordinary Mermaid fences in notes:

````markdown
```mermaid
flowchart TD
  Note --> Emanote
  Emanote --> SVG
```
````

The generated page should contain the rendered SVG directly, so diagrams work in static exports without loading Mermaid JavaScript from a CDN and without requiring network access in the reader's browser.

## Expected behavior

- Mermaid diagrams render in generated HTML and live preview.
- Static sites remain fully viewable offline.
- The existing authoring syntax stays Markdown-native: fenced code blocks with the `mermaid` class.
- Diagram rendering runs only for HTML output; it should not affect Emanote's note model, links, tags, backlinks, search metadata, or other parse-time behavior.
- Any external renderer executable needed for diagrams should come from Emanote's Nix environment, not from an undeclared user PATH dependency.
- Rendering failures should be visible to the author with a clear message, while preserving enough source context to fix the diagram.

## High-level design

Use Emanote's render-time Pandoc Lua filter phase for this feature.

Mermaid rendering is an output concern: it produces HTML/SVG and may require IO or external tooling. That makes it a render-time transform, not a parse-time transform. The note should first be parsed normally, and the Mermaid-to-SVG conversion should happen only when Emanote is producing HTML.

Prefer a reusable Pandoc Lua filter approach, such as the `pandoc-ext/diagram` direction, over a Mermaid-specific Emanote renderer. Emanote can document and demo the filter configuration needed for offline Mermaid, while keeping the rendering model aligned with the general Lua filter system.

Executable discovery should be explicit and reproducible: renderer binaries required by the filter should be supplied through Nix and resolved through `staticWhich`, so Emanote does not accidentally depend on whatever happens to be installed on the host.

## Related work

- [#721](https://github.com/srid/emanote/issues/721) tracks the broader Pandoc Lua filter roadmap.
- [#731](https://github.com/srid/emanote/pull/731) added render-time HTML Lua filters, which is the right phase for this feature.

### PR DESCRIPTION
**Pull request by srid**
**Mermaid code blocks now render to inline SVG at build time** — the generated site no longer needs the CDN-hosted `mermaid.js` to display diagrams and works fully offline. A new `Emanote.Pandoc.Mermaid` module walks every parsed note's AST and replaces ```` ```mermaid ```` blocks with the SVG produced by `mmdc` (mermaid-cli). The `mmdc` binary path is baked in at compile time via `staticWhich` — same convention as `tailwindBin`/`storkBin` — so the dep is enforced by the Nix flake rather than discovered at runtime. _No fallback by environment: server-side rendering is the default, full stop._ The retained `js.mermaid` snippet stays documented as an opt-in alternative for users who explicitly want client-side CDN rendering.

The transform runs inside `parseNote`'s `WriterT` block, so per-diagram render failures join the same per-note error stream as parser failures and surface in the document-top error banner. A failing diagram is also rendered inline as a visible `Mermaid rendering failed:` message above its preserved source — readers see both signals at the right granularity.

> **Upstream fix:** The mermaid SVG path tripped a long-standing bug in `heist-extra`'s `rawNode`: xmlhtml errors with `div cannot contain text looking like its end tag` whenever a raw HTML blob contains a literal `</div>` (which mermaid's `<foreignObject>`-based labels always do). Fixed in [srid/heist-extra#13](https://github.com/srid/heist-extra/pull/13) by switching the wrapper to a unique `<rawhtml>` element with `display: contents`. The flake input is pinned to that branch until the PR lands on master.

> **Deferred:** The structural review (Hickey/Lowy lenses on Opus) flagged three follow-ups worth tracking — config knob over rendering mode (#653), content-addressed cache to avoid re-running `mmdc` on every parse (#654), and widening the renderer interface so this transform can move from parse-time to render-time (#655).

Closes #625, closes #119.

### Try it locally

```sh
nix run github:srid/emanote/polite-claim
```


srid:

Reviewers ran on **opus** per `--review-model=opus`.

| # | Lens   | Finding                                              | Disposition       |
|---|--------|------------------------------------------------------|-------------------|
| 1 | Hickey | Fragmented error channel (logger vs WriterT errs)    | Fixed in this PR  |
| 2 | Hickey | Parse-time IO complects parsing with rendering       | Deferred [#655](https://github.com/srid/emanote/issues/655) |
| 3 | Hickey | "Always-on if PATH" interleaves environment with semantics | Deferred [#653](https://github.com/srid/emanote/issues/653) |
| 4 | Hickey | No caching: render coupled to build                  | Deferred [#654](https://github.com/srid/emanote/issues/654) |
| 5 | Hickey | `Div .mermaid-error` packs three concerns; error hidden in HTML comment | Fixed in this PR |
| 6 | Hickey | `stripXmlPrologue` silently truncates malformed input | Fixed in this PR |
| 7 | Hickey | Implicit pipeline order across md/org branches       | Deferred (cosmetic) |
| 8 | Lowy   | Per-note opt-out unencapsulated; sequence volatility named but rigid | Deferred [#653](https://github.com/srid/emanote/issues/653) |
| 9 | Lowy   | XML prologue stripping narrow (only `<?xml ?>`/`<!DOCTYPE>`) | Fixed in this PR |
| 10 | Lowy   | Error message escaping incomplete                    | No-op (subsumed by #5: visible message replaces HTML comment) |
| 11 | Lowy   | Pipeline placement is implicit at call site          | Fixed in this PR  |

### Hickey rationale

The module itself is structurally clean; the **wiring** is where ease beat simplicity. The decisive complect was the parse-time hook (#2): SVG generation is "turn AST into a presentation artifact" and belongs at render time, but the existing `PandocBlockRenderer` returns `Splice Identity` and can't perform IO. The author chose parse-time as the at-hand fold rather than widening the render interface — a real ease-vs-simplicity trade. Fixing it in this PR would mean a separate refactor of the renderer machinery; deferred to #655.

The dual error channel (#1, fixed) and the hidden HTML-comment diagnostic (#5, fixed) were the two complects that could be unwound without architectural change. Errors now flow through a single `WriterT [Text] m` channel, and a failed diagram displays a visible `Para` above the preserved source instead of a comment readers never see. The malformed-input silent truncate in `stripXmlPrologue` (#6) was a classic "parser dressed as `dropWhile`" — replaced with `T.breakOn` returning the input unchanged on no-terminator.

The PATH-driven on/off behaviour (#3) is the remaining structural concern: the same `.md` produces different output on hosts with vs. without `mmdc`. A proper fold needs an explicit `mermaid.renderer: server | client | off` knob with `server` erroring loudly when `mmdc` is absent. Deferred to #653 because it's a config-feature design question worth handling on its own.

### Lowy rationale

Volatility map: **mmdc CLI surface** (lives in `runMmdc` — correctly encapsulated as activity volatility), **SVG format quirks** (`stripXmlPrologue` — pure, narrow, now extended to handle arbitrary `<?...?>` PIs per #9), **error envelope** (`renderBlock`'s `Div .mermaid-error` — correctly encapsulated). The boundary is sound and named after the right axis.

The two correctness gaps were patched in this PR: prologue robustness now leaves malformed prefixes untouched and consumes any leading processing instruction; the implicit ordering contract at the call site (#11) is now stated explicitly so a future reader doesn't have to reverse-engineer "mermaid runs after Lua filters because user filters get first crack at mermaid blocks." The error-escape concern (#10) became moot once #5's visible message replaced the HTML-comment diagnostic — there is no comment to break out of.

The remaining unfixed volatility is sequence (when does rendering happen) and per-site/per-note opt-out, both packaged into #653. Caching as a separate volatility axis lives in #654 — explicitly _not_ inside `transformMermaidBlocks` if added later, since cache invalidation is a different rate of change from the AST traversal it would speed up.

srid:

| research | ✓ | 6m 44s | Mapped renderer pipeline; hook point at `Note.hs:parseNote` with `WriterT [Text]` |
| branch | ✓ | 0s | On feature branch `polite-claim` |
| implement | ✓ | 4m 17s | New `Emanote.Pandoc.Mermaid`, wired in `parseNote` (md+org), Nix flake + tests |
| check | ✓ | 1m 55s | `cabal build all` clean; `pkgs.mermaid-cli` v11.12.0 |
| docs | ✓ | 43s | Updated `docs/tips/js/mermaid.md` |
| fmt | ✓ | 27s | fourmolu/cabal-fmt/hlint/nixpkgs-fmt all clean |
| commit | ✓ | 1m 46s | Feature commit pushed |
| hickey+lowy | ✓ | 9m 59s | Opus sub-agents; 4 Fix-in-PR commits, 3 deferred to #653 #654 #655 |
| police | ✓ | 9m 47s | rules+fact-check clean; 4 elegance refinements as separate commits |
| test | ✓ | 10s | 51 examples, 0 failures |
| create-pr | ✓ | 2m 15s | Draft PR opened with Hickey/Lowy analysis |
| ci | ✓ | 5m 48s | flake-parts-docs (1m48s), e2e live (4m27s), e2e static (4m28s) |
| **Total** | | **44m 12s** | |

**Slowest step**: `hickey+lowy` (9m 59s)

### Optimization suggestions

- **Review steps dominated** (`hickey+lowy` 10m + `police` 10m ≈ 45% of run): `--review-model=opus` was the right call given the structural depth surfaced (parse-time-IO complect, dual error channel, env-driven semantics), but on smaller diffs `--review-model=sonnet` would shave ~6-8 minutes with comparable rule/fact-check coverage. Reserve opus for diffs that cross module boundaries.
- **`research` at 6m 44s**: two Explore subagents needed before the hook point was clear. Pre-reading `Emanote.Pandoc.Renderer` and `Emanote.Model.Note` before invoking `/do` would have collapsed this to one subagent or zero.
- **CI at 5m 48s** is dominated by playwright browser install + e2e — caching is already configured. Re-runs on followups should hit cache and be ~3 minutes.
- **`test` at 10s** confirms the unit suite is fast enough to run on every iteration; no incremental savings to be had there.

Workflow completed at $(date -u +%Y-%m-%dT%H:%M:%SZ).

Issue #632, opened 2026-04-21
Modernize UI/UX (tracking: post-Tailwind-v4 work)


srid:
Tracking issue for the visual + UX modernization work that follows the Tailwind v4 migration. Split out deliberately so the v4 migration lands and stabilizes in isolation — regressions there should be attributable to the tokens change, not bundled with redesign.

### 1. Visual refresh

- Spacing, typography, sidebar density pass on the existing layouts in `emanote/default/templates/`.
- Native `<details>`-based collapsible sidebar, replacing the `toggleHidden()` inline script in `components/breadcrumbs.tpl:41-45` and the Heist-conditional expand/collapse in `components/sidebar-tree.tpl:83` (existing TODO). This is genuine JS elimination.
- TOC scroll tracking (`components/toc.tpl:21-73`, existing TODO "We should get rid of JavaScript!" — issue #520): swap `window.onscroll` + `offsetTop` arithmetic for `IntersectionObserver`. This improves the JS, it doesn't eliminate it — see section 3 below for the module boundary.
- Proper mobile sidebar drawer (overlay, scroll-lock, focus trap). Current mobile UX is a class-toggle, not really a drawer.
- Icon partial: consolidate the copy-pasted inline heroicon SVGs in `sidebar.tpl`, `breadcrumbs.tpl`, `toc.tpl`, etc. into a single reusable Heist partial.

### 4. UX features

- Command palette (Ctrl+K already opens Stork search; extend to actions: jump-to-tag, jump-to-recent, toggle sidebar, toggle theme).
- Keyboard navigation through the sidebar route tree (arrow keys, Enter to open).
- Recently-viewed notes (localStorage), surfaced in sidebar or command palette.
- Better empty states (no results, empty folder, no backlinks).
- Proper 404 page design (currently uses `special/error.tpl`).

### 3. JS module boundary (blocks section 4)

**Decide before section 4 starts, not during.** Section 4 features need non-trivial JS. To avoid proliferating inline `<script>` blocks that accrete across templates, commit to a module boundary upfront:

- All interactive JS lives in one bundled file loaded once from `_emanote-static/`, not inlined across templates.
- Templates expose state via `data-*` attributes; JS reads them. No Heist-spliced JS values.
- Library choice: Alpine.js (declarative, minimal, ~15kb) vs. vanilla Web Components (zero deps, more boilerplate) vs. plain vanilla with a small helper module. Decide as part of scoping section 4.

[...]

PR #663, opened 2026-04-25
Move interactive JS to ES modules with live-server cache busting

### LINKED ISSUE #643: Decide JS module boundary before the cruft accretes further
Promoted out of [#632](https://github.com/srid/emanote/issues/632) (section 3, _JS module boundary_) so it can be scheduled on its own — the tracking issue gates section 4 on this decision, but the decision also blocks further interactive JS _anywhere_, and the list of places that need it keeps growing.

**Every new interactive feature lands as another inline IIFE in a Heist template**, and the count is now five. Left alone, section 4 adds command-palette, keyboard nav, and recently-viewed, each naturally pulling the same gravity.

Current inventory of inline interactive JS, by file:

| Template | Behavior | Rough size |
|---|---|---|
| `emanote/default/templates/components/stork/stork-search-head.tpl:17-99` | Stork search controller — modal open/close, keyboard shortcuts, index loader | ~80 lines |
| `emanote/default/templates/components/styles.tpl:289-305` | Theme toggle — persists in localStorage, publishes `window.emanote.theme.toggle` | ~15 lines |
| `emanote/default/templates/components/styles.tpl:307-344` | Code-copy-button injection — creates a button per `<pre>`, clipboard write + 2s revert | ~40 lines |
| `emanote/default/templates/components/toc.tpl:21-73` | TOC scroll tracker — `window.onscroll` + `offsetTop` arithmetic; issue [#520](https://github.com/srid/emanote/issues/520) already wants `IntersectionObserver` | ~50 lines |
| `emanote/default/templates/components/breadcrumbs.tpl:41-45` | `toggleHidden()` sidebar class-toggle | ~5 lines |
| `emanote/default/templates/components/footnote-popup.tpl` | Popup footnotes — popover lifecycle, scope-aware target resolution, viewport repositioning, live-server reconnection guards (landed in [#642](https://github.com/srid/emanote/pull/642)) | ~115 lines |

That last row is new in PR #642 and is exactly the pattern [#632's section 3](https://github.com/srid/emanote/issues/632) names as the failure mode: a self-contained IIFE in a `.tpl` that reaches into DOM via `data-*` attributes, colocated with the CSS it styles. It is _individually_ defensible (the attribute contract is deliberate, the lifecycle is self-contained) but it is another data point on the same trajectory — six files now share a pattern nobody has signed off as the house style.

### What needs to be decided

- **Module shape.** One bundled file loaded once from `_emanote-static/` vs. module scripts co-located with templates vs. Web Components. The tracking issue mentions Alpine.js (~15kb, declarative), Web Components (zero-dep, boilerplate), or vanilla-with-a-helper. Pick one and commit before the next IIFE gets written.
- **State-passing contract.** `data-*` attributes only, no Heist-spliced JS values. This is already de-facto the convention (popup footnotes do exactly this), but it should be documented.
- **Where it runs.** Loaded once in `base.tpl`, module-level IIFEs that idempotently attach on `DOMContentLoaded` _and_ are resilient to Emanote's live-server DOM patches (the footnote popup already learned this the hard way — `ensurePopover` / `findTarget` both do `isConnected` reconciliation).
- **Migration plan.** Consolidation order for the six existing scripts. Some are trivial (breadcrumbs toggle is 5 lines and will evaporate into `<details>` in section 1 anyway); others (stork, TOC tracker) are non-trivial rewrites that want their own PRs.

### Why now

Section 4 of #632 can't start without this. Section 1 is actively adding more inline scripts (#642 added one; native `<details>` in section 1 will remove one but the trend is clearly upward). Each new IIFE that lands before the boundary is decided is another migration target later.

Blocks [#632 section 4](https://github.com/srid/emanote/issues/632). Related: [#520](https://github.com/srid/emanote/issues/520) (TOC scroll tracking), [#642](https://github.com/srid/emanote/pull/642) (the most recent IIFE).

### LINKED ISSUE #667: JS bundle survives Ema morph navigation poorly: TOC scroll-spy dies, importmap warnings fire
**Reproduction:** open the dev server (`emanote run`), load `/`, click into Guide, then click into Markdown. Console emits five warnings:

```
An import map rule for specifier '@emanote/footnote-popup' was removed, as it conflicted with already resolved module specifiers.
An import map rule for specifier '@emanote/theme-toggle' was removed, …
An import map rule for specifier '@emanote/morph' was removed, …
An import map rule for specifier '@emanote/toc-spy' was removed, …
An import map rule for specifier '@emanote/code-copy' was removed, …
```

Then scroll the Markdown page. The TOC is rendered with all 20 links, but none of them ever gets the `toc-item-active` class — scrolling does nothing visually. Reaching the same page via a fresh load (browser address-bar navigate, or hard reload) instead of via in-app clicks works correctly.

Verified end-to-end via chrome-devtools MCP against `http://127.0.0.1:9010`:

| Path | TOC links | Active after scroll to `#tables` |
|---|---|---|
| Fresh load `/markdown` (page.goto) | 20 | `markdown#tables` ✓ |
| Morph nav `/` → `/guide` → `/markdown` | 20 | none ✗ |

### Two stacked problems, one root cause

The cause is the same in both: **the JS bundle's module-level state is per-document-load, and Ema's morph-navigation does not reload the document**.

1. **importmap conflict warnings.** The new HTML for `/markdown` carries a `<script type=\"importmap\">` that maps `@emanote/morph` → `morph.js?t=…` etc. By the time it lands in the morphed `<head>`, the bare specifiers have already been resolved on the previous page. Per the HTML spec, an importmap can only ADD specifiers once any module has been resolved — conflicting entries are silently removed. So the new URLs (potentially with new `?t=` values after a file edit) are ignored. Live-mode cache busting therefore breaks the moment you navigate without a full reload — you have to hard-refresh to see edits to `morph.js` etc. when the page change came via a sidebar click.

2. **toc-spy doesn't re-initialize.** `toc-spy.js`'s top-level body wraps everything in `ready(() => {…})`, runs once at first DOMContentLoaded, captures `document.querySelectorAll('a.--ema-toc')` into a closure, and registers an IntersectionObserver against those exact nodes. After morph the old TOC links are gone (replaced); the observer is left holding refs to detached nodes, and the new TOC links have no observer at all. `pickActive()` never runs for the new page → no active link, no updates on scroll.

`footnote-popup.js` survives this by design — single delegated listener on `document`, lazy `isConnected` re-resolution. `code-copy.js` survives via the `morph.js` MutationObserver (a new `<pre>` from any source gets wired). `theme-toggle.js` is just a global; doesn't care. `toc-spy.js` is the odd one out — it does a one-shot scan and trusts the closure.

### Fix space

The two problems want different remedies:

- **For the importmap warnings:** options are (a) mark the importmap with `im-preserve` so idiomorph keeps the original (the spec doesn't let us replace it anyway, so re-emitting it on morph is pure noise + breaks cache-busting via morph); or (b) emit the importmap once via Ema's preserved-head mechanism. Either way, **on file edit the morph path can't pick up the change** — only a full reload can — so the cache-busting story needs an explicit caveat (or a force-reload trigger from the live-server WS when a JS file changes).

- **For toc-spy:** rewrite so it doesn't rely on a one-shot closure. Either listen to the `EMABeforeMorphDOM` event (or the equivalent after-morph hook) and re-scan, or use `morph.js`'s `onElement` primitive (extend it to take a teardown callback so old observers can be discarded). The same audit applies to any future behavior — the de-facto convention should be \"survives morph by design\", with the test that proves it (currently every e2e scenario uses `page.goto`, which is fresh-load only — see #663 PR description for the test-infrastructure gap).

### Adjacent: 404s on sourcemap paths

Server logs show repeated 404s like:

```
[Info#ema.http] GET sm/e2c774c727d3cbe1cc4e8eced2ee28acf2def82c050a4341743a7e3315aa70ae.map as Right (Just 404: sm/…)
```

These are Tailwind v4 CDN sourcemap requests against content-hashed paths the live server doesn't serve. Probably orthogonal — separate issue if it warrants one.

Related: #663 (the JS-module-boundary PR that introduced the importmap), #643 (the architecture decision).

srid:
**Site-authored interactive behaviors no longer live as inline `<script>` IIFEs scattered across Heist templates.** Four small ES modules under `_emanote-static/js/` (`theme-toggle`, `code-copy`, `toc-spy`, `footnote-popup`) replace the per-template IIFEs, plus a `morph.js` helper providing `ready()` and `onElement(selector, fn)` (MutationObserver-backed). Closes #643. *Side effect*: code-copy buttons now wire onto `<pre>` blocks added by live-server DOM patches, not only those present at first paint.

Live-mode cache busting works **per file** via an inline `<script type="importmap">` mapping bare specifiers (`@emanote/morph` etc.) to URLs through `siteRouteUrl`, which appends `?t=<mtime>`. Plain ES module imports (`import './morph.js'`) would have stripped the query during URL resolution and stale-cached transitive deps; the bare-specifier indirection is what makes per-file invalidation propagate. The importmap itself is tagged `im-preserve` + `data-ema-skip` so it survives Ema's morph navigation — without that, the second importmap in the morphed `<head>` hit the HTML spec's *already-resolved specifier* rule and emitted a console warning per entry. The `toc-spy` module captures DOM refs in a closure, so it now also subscribes to Ema's `EMAHotReload` event to tear down + re-init on each morph (closes #667).

Static mode also gained a fix: the production Tailwind compile now adds `_emanote-static/js` as an `@source`, so utility classes applied by JS at runtime (`POPOVER_CLASS`, `BODY_CLASS` in `footnote-popup.js`) are visible to the scanner and don't drop out of `tailwind.css`. Pre-fix, the popup rendered as plain text inside a thin border on `emanote gen` output. Also strips Tailwind's `//# sourceMappingURL=/sm/<hash>.map` trailer to silence the per-pageload 404 in dev logs.

> **Depends on srid/ema#181** (merged) — `window.ema.ready` Promise so the morph-nav e2e step can await the WS handshake without racing.
>
> **Out of scope, kept inline by design**: FOUC theme applier in `base.tpl` (pre-paint, can't defer), Stork controller (`onclick="window.emanote.stork.toggleSearch()"` call sites span four templates — separate refactor), breadcrumbs `toggleHidden` (5 lines; section 1 of #632 deletes via native `<details>`).
>
> **Known limitation, documented**: editing a *bare-imported* JS module mid-session needs a hard reload to pick up the change via morph nav (the preserved importmap holds the original URLs). `main.js` itself still cache-busts via morph since it's loaded directly. Issue #666 tracks the broader layer-URL cache-bust gap (CSS, fonts, Stork's own assets).

E2e: 16 scenarios, live=16/16, static=15+1-skipped (`@morph` tag gates morph-nav for live only). New scenarios cover theme toggle, code-copy initial wiring, TOC scroll-spy, popup body styling (catches the Tailwind class-drop regression class), and morph-nav of toc-spy. Module split into `Emanote.View.JsBundle` mirrors the existing `Emanote.View.Tailwind` shape.

### Try it locally

```sh
nix run github:srid/emanote/js-module-boundary -- -L $YOUR_NOTEBOOK run
```


srid:

| 1 | Hickey | Storage key duplicated between `base.tpl` FOUC and `theme-toggle.js` | Fixed in this PR  |
| 2 | Hickey | Silent `localStorage.setItem` failure                                | Fixed in this PR  |
| 3 | Hickey | `CSS.escape` not gated by the popover feature-detect                 | Fixed in this PR  |
| 4 | Hickey | `flash()` timer race on rapid re-clicks of the copy button           | Fixed in this PR  |
| 5 | Lowy   | `window.emanote` namespace ownership is implicit                     | Deferred          |
| 6 | Lowy   | Storage-key shared volatility (FOUC inline + theme-toggle module)    | Deferred          |

### Hickey rationale

Four findings, all addressable in this PR. The storage-key duplication is structurally unavoidable without a Heist splice on a JS literal (the FOUC script must run synchronously before any module loader executes), so the fix is the cheapest hardening that helps: a bidirectional pointer comment so a future edit on either side finds the other. The other three are real correctness bugs — bare `try { localStorage.setItem(…) } catch (e) {}` masking persistence failures (now `console.warn`), a `setTimeout` on the copy-button flash that gets overwritten by the next click within the 2 s window (now per-button cancel via `WeakMap`), and a feature-detect guard that checked for `Popover` but not `CSS.escape` even though `findTarget` calls the latter (now both predicates).

### Lowy rationale

Two findings, both deferred. The implementation matches the design sketch on every structural axis: `morph.js` correctly encapsulates the Ema-framework dependency axis (live-reload resilience), `theme-toggle` and the FOUC inline correctly carve along the paint-blocking-vs-interactive axis, and Stork is cleanly deferred (no half-migration markers). The two open items are framing concerns that don't require a fix here. The `window.emanote` namespace contract is currently maintained by defensive `|| {}` guards in both stork and theme-toggle; documenting or consolidating it makes sense once #632 section 4 adds more entries (command palette, kbd nav, recently-viewed) — premature now. The storage-key duplication concurs with Hickey's finding #1 — both reviewers agreed structural deduplication needs out-of-scope work; the bidirectional comment is the in-scope mitigation.

srid:

| research | ✓ | 1m 50s | research map produced; plan articulated with file:line citations |
| branch | ✓ | 6s | on branch `js-module-boundary` tracking `origin/master` |
| implement | ✓ | 3m 31s | 6 JS modules created; 4 templates pruned; cabal `data-files` updated |
| check | ✓ | 44s | `cabal build all` succeeded |
| docs | ✓ | 46s | CHANGELOG.md updated under UI revamp |
| fmt | ✓ | 12s | pre-commit clean (cabal-fmt, fourmolu, hlint, nixpkgs-fmt) |
| commit | ✓ | 22s | primary feature commit pushed |
| hickey+lowy | ✓ | 5m 48s | hickey: 4 findings, 4 fixed; lowy: 2 findings, both deferred with rationale |
| police | ✓ | 3m 58s | rules clean; fact-check clean; elegance: 1 fix (WeakMap for flash timers) |
| test | ✓ | 1m 23s | e2e-static: 12/12 scenarios pass (incl. 3 footnote-popup click scenarios) |
| create-pr | ✓ | 1m 5s | draft PR #663 + hickey/lowy analysis comment |
| ci | ✓ | 1m 30s | `vira ci` succeeded; 5 packages × 2 platforms; gh-signoff posted |
| **Total** | | **21m 38s** | |

**Slowest step**: `hickey+lowy` (5m 48s)

### Optimization suggestions

- **`hickey+lowy` dominated** (27% of wall-clock). The two parallel sub-agent reviews ran on `sonnet` against a 440-line diff — opus would have been slower. The four small follow-up commits (one per finding) compounded the time; rolling them into a single "address review" commit would save ~30 s but lose the per-finding history.
- **`police` was second slowest** (18%). Three parallel review agents in the elegance pass produced one actionable finding out of ten total suggestions; many were over-abstraction (try/catch helper, class-name constants for 2-site references). For PRs of this size, a quicker manual elegance pass might match the hit rate.
- **`implement` (3m 31s)** is the irreducible cost — six new files plus four template edits. Splitting this PR (e.g., morph + code-copy first, then footnote-popup + toc-spy) would reduce per-PR review overhead but add per-PR fixed costs.
- For re-runs after CI failures, use `--from ci-only` to skip everything before the CI step.

Workflow completed at 2026-04-25T23:19Z.

srid:
## [Hickey/Lowy](https://kolu.dev/blog/hickey-lowy/) Analysis (resumed run)

| # | Lens   | Finding                                                                  | Disposition       |
|---|--------|--------------------------------------------------------------------------|-------------------|
| 1 | Hickey | `sentinelKey()` selector-collision footgun in `morph.js`                 | Fixed in this PR  |
| 2 | Hickey | `EMAHotReload` event name leaks into behavior code                       | Fixed in this PR  |
| 3 | Hickey | `@morph` tag convention unenforced — silent 60s timeout on missing tag   | Fixed in this PR  |
| 4 | Hickey | Dual manifest (main.js imports + `emanoteJsModuleNames`) drift risk      | Fixed in this PR  |
| 5 | Lowy   | JsBundle.hs conflates volatile (cache-busting) and stable (scaffolding)  | Fixed in this PR  |
| 6 | Lowy   | Tailwind.hs absorbs JS-dir layout knowledge                              | Fixed in this PR (closes #670) |
| 7 | Lowy   | Dual manifest as missing single source of truth                          | Fixed in this PR (closes #669) |
| 8 | Lowy   | Morph-survival convention undocumented across 4 strategies               | Fixed in this PR (closes #671) |

### Hickey rationale

Four findings, all addressed.

(1) `sentinelKey()` derived a dataset key from the selector by character-substitution — `sup[data-foo]` and `sup[data_foo]` would collide silently. Replaced with a per-registration counter; the sentinel is opaque, collision-free.

(2) `EMAHotReload` was hardcoded in `toc-spy.js` and would have been hardcoded in every future morph-aware behavior. Wrapped in `morph.js`'s new `onMorph(fn)` export; the event name is now a one-place internal detail.

(3) A scenario using `I navigate via Ema to ...` without the `@morph` tag would silently run in static mode (no WS, `window.ema` undefined) and time out at the step's 60s ceiling. New `Before` hook scans steps + tags and throws fast with a pointer to the right tag.

(4) `main.js`'s import list and `JsBundle.hs`'s `emanoteJsModuleNames` were two views of the same thing. Initial fix was a cross-pointer comment; full fix landed under Lowy #7 below.

### Lowy rationale

Four findings, all addressed (the `Defer` column is empty — per request, every finding shipped here).

(5) `JsBundle.hs` was named after what it does (JS bundling) rather than what changes (the cache-busting protocol). Trimmed the docstring to one paragraph that names the volatility boundary explicitly: only `jsUrl` / `importmapUrl` are expected to evolve; the manifest, importmap shape, and HTML rendering are W3C-spec-stable scaffolding co-located for locality, not for shared change rate.

(6) The `_emanote-static/js` source path lived inside `compileTailwindCss`. Extracted to a tiny `Emanote.Static.Sources` module that exports `jsScanPath`. Tailwind reads from it; future scanners (lints, asset pipelines) read from the same registry. Closes #670.

(7) The dual-manifest hand-bookkeeping is gone: `main.js` now parses the inline `<script type="importmap">` at runtime and dynamically imports each bare specifier. The Haskell `emanoteJsModuleNames` is the single source of truth; the loader has no second list. Adding a behavior is a one-line Haskell change. Browser dedupes module-loads by URL, so behaviors that statically import each other still hit the same cached instance. Closes #669.

(8) `morph.js`'s header now names and characterizes all four morph-survival strategies (stateless global, event delegation + lazy resolution, per-element wire-on-appear, re-init on morph) with concrete pickers. The future `morph.withSurvival(strategy, init)` consolidation stays speculative — only worth it if more behaviors land. Closes #671.

E2e remains 17/17 in live, 16+1-skipped in static.

srid:
## [`/do`](https://github.com/srid/agency) results (resumed run, post-implement entry)

| sync → commit | ✓ | (already done) | merge of origin/master clean; check/docs/fmt/commit all green from prior run |
| hickey+lowy | ✓ | 9m 27s | 4 hickey + 4 lowy findings; **all 8 fixed in this PR** (no defers, per request) |
| police | ✓ | 8m 22s | rules + fact-check clean; elegance: 2 small fixes + 3 previously-deferred Lowy items shipped (#669/#670/#671 closed) |
| test | ✓ | 28s | e2e static 16+1-skipped/17; e2e live 17/17 (incl. morph-nav scenario) |
| create-pr | ✓ | 1m 4s | resumed-run hickey/lowy comment posted with 8-row findings ledger |
| ci | ✓ | 1m 32s | `vira ci` passed on `cc081781`; 5 packages × 2 platforms; signoffs landed |
| **Total** | | **21m 31s** | |

**Slowest step**: `hickey+lowy` (9m 27s) — bigger diff this time (~900 LOC across 23 files vs. the original 6-file extraction), more for both reviewers to chew on.

### Optimization suggestions

- **`hickey+lowy` and `police` together = 82% of wall-clock.** Both ran two/three review sub-agents in parallel; the cost is mostly the agent latency itself, not anything fixable on our side. For incremental PRs that don't grow much between resumed runs, a smaller scope hint (e.g. "review only commits since `<sha>`") would shorten both.
- **`test` was fast (28s)** because no Tailwind compile happened — both modes already have cached state. Cold runs are ~2 min.
- **No CI retries** — the cumulative diff built clean on first attempt across both platforms.

Workflow completed at 2026-04-26T16:44Z. PR is ready for review.

PR #664, opened 2026-04-25
Resolve markdown links to .xml assets when no feed shadows them (#547)

### LINKED ISSUE #547: Can't link to an xml file
**Describe the bug**
Whenever I try to link to an `.xml` file using a regular markdown link (`[Test](./test.xml)`), Emanote displays an invalid link. Using the alternative syntax with double brackets works (`[[test.xml]]`).

It doesn't seem to happen with files with other extensions (`.json` or `.txt`, for example).

**To Reproduce**
1. Create a file named `text.xml`.
2. Create a markdown file which links to it using the regular Markdown syntax - `[Test](./test.xml)`.
3. Observe that the rendered link in the browser looks as if the file didn't exist.

**Expected behavior**
Link should be rendered correctly, similarly to other file types (`.json`) or the double-bracket syntax (`[[test.xml]]`).

**Screenshots**

The screenshot displays the following markdown content:
```markdown
# Test

- [Test](./test.xml)
- [Test](test.xml)
- [[test.xml]]
```
image

 - Browser: Firefox
 - Version: `130.0`
 - OS: MacOS 14.6.1 (23G93)

srid:
**A regular markdown link to a static `.xml` asset — `[Test](./test.xml)` — now resolves to the file** instead of rendering as broken. Previously every `.xml` URL was claimed exclusively by the Atom-feed slot in `mkModelRouteFromFilePath`, so when no matching feed-enabled note existed the resolver gave up at the note-lookup miss and never tried the static-file interpretation.

The fix lives in the **parser layer**, not patched into the resolver. `mkModelRouteCandidates` now returns the full list of plausible interpretations for a URL, and `URTResource` carries a `NonEmpty ModelRoute` of those candidates. Resolution is a pure first-hit-wins fold over the list — _the resolver no longer carries any extension-specific knowledge_. Backlink lookups round-trip through the same candidate builder so the URT shape matches what parse stored in the rel index.

Existing behaviour is preserved: feed links (`[atom](feed.xml)` against a feed-enabled note) hit the first candidate and resolve to the feed exactly as before; wikilinks (`[[test.xml]]`) take a different code path entirely and were never affected.

The **missing-link page** also got a fix from the same root cause: navigating to a missing `/foo.xml` used to suggest `foo.xml.md` / `foo.xml.org`, which is nonsensical. The hint now classifies the URL extension — `.xml` suggests creating the static file _or_ a feed-enabled `foo.md`; other non-LML extensions (`.css`, `.js`, …) suggest the static file alone; LML/HTML/no-extension URLs keep the existing `foo.md` / `foo.org` suggestion.

A Cucumber smoke scenario (`xmllink.md` → `test.xml`) guards the regression in both live and static modes; existing unit tests (`RelSpec.parseUnresolvedRelTarget`) still pass without modification beyond the new candidate-list shape.

Closes #547.

### Try it locally

```sh
nix build github:srid/emanote/fix-547-xml-link
```

### DIFF
--- a/emanote/CHANGELOG.md
+++ b/emanote/CHANGELOG.md
@@ -20,6 +20,7 @@
 **Bug fixes**
 
 - Atom feed: a feed query that matches no notes no longer crashes the build; an empty-but-valid Atom document is emitted instead. Configuration errors (missing/invalid query block, missing `page.siteUrl`) still fail loudly ([#490](https://github.com/srid/emanote/issues/490), [#650](https://github.com/srid/emanote/pull/650))
+- Markdown links to a static `.xml` asset (e.g. `[Test](./test.xml)`) now resolve to the file. Previously a `.xml` URL was always interpreted as the Atom feed of a same-named note, leaving asset links broken when no such feed-enabled note existed (closes [#547](https://github.com/srid/emanote/issues/547))
 - Resolve relative URLs inside `<dir>/index.md` against `<dir>/` instead of its parent ([#651](https://github.com/srid/emanote/pull/651), closes [#608](https://github.com/srid/emanote/issues/608))
 - Raw HTML blocks containing a literal `</div>` no longer crash the renderer with `div cannot contain text looking like its end tag` (closes [#119](https://github.com/srid/emanote/issues/119)). Fixed upstream in [srid/heist-extra#13](https://github.com/srid/heist-extra/pull/13) by switching the raw-HTML wrapper to a unique `<rawhtml>` element with `display: contents`.
 
--- a/emanote/CHANGELOG.md
+++ b/emanote/CHANGELOG.md
@@ -20,7 +20,7 @@
 **Bug fixes**
 
 - Atom feed: a feed query that matches no notes no longer crashes the build; an empty-but-valid Atom document is emitted instead. Configuration errors (missing/invalid query block, missing `page.siteUrl`) still fail loudly ([#490](https://github.com/srid/emanote/issues/490), [#650](https://github.com/srid/emanote/pull/650))
-- Markdown links to a static `.xml` asset (e.g. `[Test](./test.xml)`) now resolve to the file. Previously a `.xml` URL was always interpreted as the Atom feed of a same-named note, leaving asset links broken when no such feed-enabled note existed (closes [#547](https://github.com/srid/emanote/issues/547))
+- Markdown links to a static `.xml` asset (e.g. `[Test](./test.xml)`) now resolve to the file. Previously a `.xml` URL was always interpreted as the Atom feed of a same-named note, leaving asset links broken when no such feed-enabled note existed. The missing-link page now also tailors its "you may create…" hint to the URL extension instead of always suggesting `<url>.md` / `<url>.org` (closes [#547](https://github.com/srid/emanote/issues/547))
 - Resolve relative URLs inside `<dir>/index.md` against `<dir>/` instead of its parent ([#651](https://github.com/srid/emanote/pull/651), closes [#608](https://github.com/srid/emanote/issues/608))
 - Raw HTML blocks containing a literal `</div>` no longer crash the renderer with `div cannot contain text looking like its end tag` (closes [#119](https://github.com/srid/emanote/issues/119)). Fixed upstream in [srid/heist-extra#13](https://github.com/srid/heist-extra/pull/13) by switching the raw-HTML wrapper to a unique `<rawhtml>` element with `display: contents`.
 
--- a/emanote/src/Emanote/Model/Link/Rel.hs
+++ b/emanote/src/Emanote/Model/Link/Rel.hs
@@ -44,7 +44,11 @@ data Rel = Rel
 -}
 data UnresolvedRelTarget
   = URTWikiLink (WL.WikiLinkType, WL.WikiLink)
-  | URTResource ModelRoute
+  | -- | One or more candidate `ModelRoute`s the URL could resolve to,
+    -- ordered by preference. A single URL can map to multiple kinds
+    -- (e.g. a @.xml@ URL → feed-enabled note OR static @.xml@ asset);
+    -- resolution picks the first candidate that exists in the model.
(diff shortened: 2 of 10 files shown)


srid:

| # | Lens   | Finding                                                    | Disposition       |
|---|--------|------------------------------------------------------------|-------------------|
| 1 | Hickey | `xmlStaticRouteFromLml` round-trips logic that inverts `mkLMLRouteFromFilePath` — extension-to-route-kind mapping is now split across `ModelRoute.hs` and `Resolve.hs` | Deferred — broader refactor |
| 2 | Hickey | Fix lives in the resolver; the parser layer would let `Resolve.hs` stay agnostic of `.xml`-as-static semantics | Deferred — broader refactor |
| 3 | Hickey | `xmlStaticRouteFromLml`'s `Maybe` is vacuously safe; doc comment was misleading | Fixed in this PR (commit 2c7a36b3) |
| 4 | Hickey | Test fixture and smoke scenario are minimal and not duplicative | No-op |
| 5 | Lowy   | All boundary/volatility checks pass; resolver is the correct layer (only one with model state); blast radius contained | No-op |

### Hickey rationale

The diff adds a fallback arm in `resolveModelRoute` that handles the `LMLView_Atom` slot specifically: try the note lookup, and on miss reinterpret the same route as a `.xml` static-file route via `xmlStaticRouteFromLml`. The structure is locally correct.

The structural debt — _findings #1 and #2_ — is that `Resolve.hs` now carries implicit knowledge that `.xml` URLs may secretly be static files. That knowledge belongs to the parse/build layer in `ModelRoute.hs`, where the `.xml`-means-Atom assumption originates. A cleaner refactor would have `mkModelRouteFromFilePath` (or its caller in `Rel.hs`) surface both interpretations and let the resolver pick by lookup hit. Both findings deferred — the in-PR fallback is acceptable technical debt for a targeted bug fix; the proper refactor is a separate PR.

Finding #3 was a one-line clarification: the `Maybe` return is total in practice (a non-empty `NonEmpty Slug` always encodes to a non-empty filepath that re-parses), but the original signature didn't say so. Fixed by documenting the totality on the helper.

### Lowy rationale

No volatility-boundary violations. The change tracks a concrete axis of change — `.xml` URL interpretation strategy with fallback — and encapsulates it in the only layer that can sequence two model lookups. The parser has no model state and can't make the disambiguation decision; pushing the fix there would invert the dependency direction. Future drift (new `LMLView_*` variants, new feed formats) extends the existing `case` in place; blast radius stays contained to one function.

srid:
### Update — Hickey/Lowy follow-up

The two structural findings I deferred in the earlier analysis are now applied in this PR (commit 551e9096):

| # | Lens   | Finding                                                    | Disposition (now)        |
|---|--------|------------------------------------------------------------|--------------------------|
| 1 | Hickey | `xmlStaticRouteFromLml` round-trips logic that inverts `mkLMLRouteFromFilePath` | **Fixed in this PR**     |
| 2 | Hickey | Fix lived in the resolver; parser layer should own URL-to-route-kind mapping | **Fixed in this PR**     |

`URTResource` now holds `NonEmpty ModelRoute` (the ordered candidate list); `mkModelRouteCandidates` produces it; `resolveModelRouteCandidates` consumes it as a first-hit-wins fold. The `xmlStaticRouteFromLml` helper is gone. The resolver is back to a pure single-route lookup with no extension-specific branching. Extension-to-route-kind mapping is now centralised in `Route.ModelRoute`.

srid:

| research | ✓ | 3m 41s | Identified root cause in `resolveModelRoute` (Resolve.hs:67-74); `.xml` URLs always resolve as feed-only and never fall back to static-file lookup. |
| branch | ✓ | 7s | Created branch fix-547-xml-link from master. |
| implement | ✓ | 2m 56s | Implemented Atom→StaticFile fallback in resolveModelRoute; added xmllink.md and test.xml fixtures plus smoke scenario. |
| check | ✓ | 12s | `cabal build all`: Up to date. |
| docs | ✓ | 26s | CHANGELOG entry added under 1.6.0.0 / Bug fixes. |
| fmt | ✓ | 20s | `just fmt`: all hooks passed. |
| commit | ✓ | 26s | Created commit 8d9bf0a5 with fix + e2e fixtures + CHANGELOG. |
| hickey+lowy | ✓ | 8m 36s | hickey: doc-comment fix committed (findings 1,2 deferred at the time); lowy: all no-op. _Findings 1 & 2 later **applied in this PR** (commit 551e9096) per follow-up request — see comment above._ |
| police | ✓ | 7m 2s | rules clean, fact-check clean, elegance reviewed. |
| test | — | 41s | Local nix dev shell wedged on unrelated fsnotify/text dep conflict; CI verified unit tests + e2e instead. |
| create-pr | ✓ | 1m 4s | Draft PR #664 created. |
| ci | ✓ | 9m 42s | `vira ci` passed on HEAD 551e9096; signoffs for vira/aarch64-darwin and vira/x86_64-linux. |
| **Total** | | **35m 36s** | |

**Slowest step**: `ci` (9m 42s)

### Optimization suggestions

- **CI is the bottleneck**, as expected for a Haskell+Nix repo (full nix-build of emanote + docs build + html-proofer link check). Consider `--from ci-only` for re-runs that don't change source.
- **`hickey+lowy` was unexpectedly slow (8m 36s)** — both reviewers ran in parallel but the Hickey sub-agent took ~6m 30s on its own. Hickey's depth scales with diff size; for tighter fixes, the cost is mostly fixed overhead worth absorbing.
- **`test` step skipped** because of an unrelated `nix develop` environment failure (fsnotify-0.4.1.0 ↔ text-2.1.2 constraint conflict in haskell-flake's ghc-9.10.3 setup). Worth filing a separate ticket to pin / patch the dev shell so local `just test` stays usable.
- **The follow-up refactor** (lifting candidate-list into `URTResource`) added ~10m of additional work and a separate CI cycle. If the structural concern is anticipated up-front, doing the proper layer placement on the first commit would have saved that.

Workflow completed at $(date -u +%FT%TZ).

PR #665, opened 2026-04-25
Don't crash on malformed *.yaml; surface as a banner (#285)

### LINKED ISSUE #285: Malformed YAML files break Emanote
**Describe the bug**

When starting Emanote I get the error `Unable to render template` in the browser: 

```
Union mounting notebook layers; please wait ...
Ema App threw an exception
Unable to render template '/templates/error'

Once you fix the source of the error, this page will automatically refresh.
```

The console says `[Error#System.UnionMount] Change handler exception: BadInput "NonStringKey []"` after loading/checking all my note files.

**To Reproduce**
1. Start Emanote via `emanote` or `emanote run`.
2. Go to `http://127.0.0.1:8000/`
3. Get the error (n the browser)

**Expected behavior**
Not getting the error.

**Screenshots**

image

 - Firefox 99.0.1 (64-bit)
 - NixOS unstable (Flake install)
- `emanote --version` 0.6.5.4

**Additional context**
Haven't run Emanote for a few weeks/months. This is also a new computer where I copied my old notes folder.

srid:
A YAML file with a non-string mapping key like `[]: foo` used to throw `BadInput` from `patchModel`, **killing the live server's UnionMount Dynamic and aborting `emanote gen`** before any HTML was produced. Parse outcome now lives in `SData._sdataValue :: Either Text Aeson.Value` — successes go right, errors go left, no exception. The renderer walks the same `routeInits @'R.Yaml` cascade `getEffectiveRouteMetaWith` already uses and prepends an `emanote:error` Pandoc Div, so a bad `subfolder/index.yaml` shows up as a banner on the notes under `/subfolder/*` and nowhere else.

E2E coverage: a paired `broken-285.{md,yaml}` fixture plus two scenarios — one asserts the banner *is* shown on the sibling note, the other asserts it does *not* leak onto unrelated pages.

Closes #285.


srid:

| # | Lens   | Finding                                                                | Disposition       |
|---|--------|------------------------------------------------------------------------|-------------------|
| 1 | Hickey | `pure id` complects "don't crash" with "preserve last-known-good"      | No-op             |
| 2 | Hickey | `BadInput` newtype now unused but kept in `Prelude` (public API)       | No-op             |
| 3 | Hickey | E2E step bound to fixture text (`"E2E Fixture"`) instead of invariant  | Fixed in this PR  |
| 4 | Hickey | YAML errors only in logs; consider `_modelSDataErrors` map             | Deferred [#285]   |
| 5 | Lowy   | YAML policy boundary inline-at-dispatch is correct (vs moving to SData) | No-op            |
| 6 | Lowy   | "Keep last good vs. drop on error" is an unnamed volatility axis       | No-op             |
| 7 | Lowy   | Add `_modelDataErrors` field on `ModelT` now                           | Deferred [#285]   |

### Hickey rationale

**#1 (pure id complects).** `pure id` binds two independent behaviors — "don't crash" and "preserve last-known-good" — into one return value. Hickey's prescription is to introduce a sum type (`PatchResult = PatchOk ... | PatchError Text`) so the caller decides between log-and-id, log-and-evict, or future-escalation into a model-level error map. **No-op rationale:** the current 5-line `case` plus 3-line WHY comment names the policy at its only call site; an extra type or helper is in service of the deferred error-surfacing work (#4) and is premature without that target shape decided. If/when that follow-up lands, the right move is to refactor _through_ this site, not to leave a half-built seam.

**#2 (`BadInput` dead but exported).** `Emanote.Prelude.BadInput` is now unused after the throw is gone. **No-op:** removing an exported newtype is a public API change, not a bug-fix scope. A separate cleanup PR can decide whether to delete the type entirely.

**#3 (E2E coupled to fixture text).** Fixed in commit [d1c2c811](https://github.com/srid/emanote/pull/665/commits/d1c2c811). The original `text.includes("E2E Fixture")` would silently stop guarding the bug if the fixture's H1 ever got renamed. Replaced with a direct check for the bug's surface markers (`"Ema App threw an exception"`, `"Unable to render template"`) — the inverse of what the user sees when the Dynamic dies.

**#4 (errors only in logs).** Deferred to [#285](https://github.com/srid/emanote/issues/285) per srid's comment on the issue: "I will implement the proper fix when I get to it." The CHANGELOG entry explicitly notes this as the follow-up.

### Lowy rationale

**#5 (boundary).** The inline `case Left/Right` in `Patch.hs` lives at the right boundary. Moving the policy into `Model/SData.hs` (mirroring how markdown errors live inside `Model/Note.hs`) would require `SData` to sprout a per-route error field — but the markdown analog works because `parseNote` returns a `Note` that *carries* errors via `_noteErrors`; `parseSDataCascading` returns `Aeson.Value`, which has nowhere to put one. Distorting `SData` to host error policy would widen its volatility surface unnecessarily.

**#6 (unnamed volatility axis).** Lowy's prescription was to extract `handleYamlParseError :: ... -> m (ModelEma -> ModelEma)` so the "keep last good vs. drop" decision is named at a function boundary. **No-op rationale:** for a 2-line policy at one call site, a function would be the same code rearranged behind a name — and the comment already names the decision in prose ("leave the model unchanged for this route — previously-good data stays in place"). The named seam earns its keep when the function grows, which only happens with the deferred error-map work (#7). Extract then, not now.

**#7 (`_modelDataErrors` now).** Deferred. The UI shape for surfacing yaml errors in the browser (banner? sidebar? dedicated page?) hasn't been decided, and adding a field that no consumer reads couples `ModelT`'s stability to a UI choice that doesn't exist yet. Tracked alongside #4 under the issue.

srid:
## Update — deferred items now done

Following user direction, the two findings previously marked **Deferred [#285]** are addressed in this PR:

| # | Lens   | Finding                                                                | Updated disposition |
|---|--------|------------------------------------------------------------------------|---------------------|
| 4 | Hickey | YAML errors only in logs; consider `_modelSDataErrors`                 | **Fixed in this PR** |
| 7 | Lowy   | Add `_modelDataErrors` field on `ModelT` now                           | **Fixed in this PR** |

Concretely, commit [32061c55](https://github.com/srid/emanote/pull/665/commits/32061c55) adds a `_modelDataErrors :: Map (R 'Yaml) Text` field to `ModelT`, populated on parse failure and cleared on success / file deletion. `renderLmlHtml` prepends a Pandoc `Div` with class `emanote:error` (the same class the markdown error path uses) when the map is non-empty, so the banner appears on every page while a yaml file is broken.

### Why a Map and not piggybacking on `IxSData`

`IxSet`'s value comes from multi-key indexing (`IxNote` is indexed by 8 keys); yaml errors only need single-key lookup by route, so an IxSet would be over-spec'd. Folding the error into `SData` itself would interleave "successfully parsed value" with "parse error" into one type — every consumer of `_sdataValue` would have to learn to ignore error-state entries (or rely on `Aeson.Null` as a sentinel). The Map keeps the success and failure channels separate; they have different lifetimes and different consumers.

### Why a render-time prepend rather than per-note `errorDiv`

Markdown errors live in `_noteErrors` because they're per-note — a bad markdown file affects only its own page. Yaml errors aren't per-note: one bad `*.yaml` affects every note whose route descends from that yaml route. The render-time prepend keeps the banner global without forcing the model to denormalize errors onto every potentially-affected note.

### Hickey #1 / Lowy #6 follow-up

Those two findings were marked **No-op** ("`pure id` complects" / "extract `handleYamlParseError`") on the basis that the policy was a 2-line block. With this commit the policy is now `M.modelInsertDataError r err . M.modelDeleteData r` vs `M.modelInsertData sData . M.modelDeleteDataError r` — explicit branches against named model operations. The "preserve last-known-good vs drop" axis is no longer implicit; the diff now drops bad data outright, which is the correct behavior once errors have a visible surface.

srid:

| research | ✓ | 5m 36s | Read issue #285, comments, repro, Patch.hs:100-113 throws BadInput, Note.hs:_noteErrors as analog pattern. |
| branch | ✓ | 4s | On feature branch fix-285-malformed-yaml from origin/master. |
| implement | ✓ | 1m 28s | E2E fixture broken-285.yaml + regression scenario. Patch.hs YAML branch logs instead of throwing. |
| check | ✓ | 3s | cabal build all → Up to date. |
| docs | ✓ | 21s | CHANGELOG entry under 1.6.0.0 Unreleased / Bug fixes. |
| fmt | ✓ | 25s | 4 hooks (cabal-fmt, fourmolu, hlint, nixpkgs-fmt) all Passed. |
| commit | ✓ | 16s | Primary commit 269ef921 pushed. |
| hickey+lowy | ✓ | 5m 21s | Sub-agents ran in parallel. 7 findings; #3 fixed (d1c2c811); #1+#2+#5+#6 no-op; #4+#7 initially deferred, later folded into the expanded scope. |
| police | ✓ | 4m 12s | 3 passes: rules clean; fact-check fixed (dfe349e2); elegance trimmed (8058fafc). |
| test | ✓ | 1m 21s | Unit 43/43 + E2E live 13/13 + E2E static 13/13. |
| create-pr | ✓ | 1m 24s | Draft PR #665 + hickey/lowy ledger comment posted. |
| **ci** | ✓ | **10m 48s** | vira ci signed off on 32061c55 (HEAD) for both vira/aarch64-darwin and vira/x86_64-linux. Cache disabled per branch policy. |
| done | ✓ | 4s | All steps completed. |
| **Total** | | **33m 59s** | |

**Slowest step**: `ci` (10m 48s, 32% of total)

### Optimization suggestions

- **CI ran twice** (once on the minimal-fix merge commit `906dd9d1`, once on the expanded-scope commit `32061c55`) because scope grew mid-run. ~10 minutes were burned on the first CI pass that became obsolete. _If the full scope is known upfront — surface yaml errors in the model, not just log them — naming that in the original `/do` prompt avoids the throw-away first build._
- **`research` + `hickey+lowy` together took 11 minutes** and that's about as fast as it gets — issue #285 has 13 comments and the parallel sub-agent invocation is already the optimal shape. No win to chase here.
- For follow-up changes on this branch (e.g. addressing review comments), `/do --from ci-only` skips straight to vira ci on the new HEAD — useful if a single-line fix doesn't need the full research/check/test/police pipeline re-run.
- **Cache was disabled** because this isn't `master` (per `vira.hs`). On master, the same CI pipeline benefits from `cache.nixos.asia/oss` — expect this run to land much faster as an actual master push.

Workflow completed at 2026-04-25.

emanote/src/Emanote/Model/SData.hs
```diff
@@ -16,11 +16,19 @@ import Relude
 
 {- | `S` for "structured". Refers to a per-route data file represented by Aeson
  value.  Example: /foo/bar.yaml file
+
+ An `SData` represents the outcome of parsing the yaml file at its route:
+ a successful parse populates `_sdataValue`; a failed parse leaves it
+ `Aeson.Null` and records the message in `_sdataError`. Modeling the
```
srid: wtf, use `Either Aeson.Value Text` then.

srid: Done in 599d8523. Went with `Either Text Aeson.Value` (error-on-Left, the conventional order) — `parseSDataCascading` is now a one-liner: `SData (mergeAesons <$> traverse parseOne bs) r`. Holler if you actually wanted the flipped `Either Aeson.Value Text`.

Issue #411, opened 2023-01-17
High memory usage of browser tabs


dpaetzel:
**Describe the bug**
I'm not sure whether it's a bug but my emanote browser tabs seem to use a lot of RAM; 1GB and more each (as displayed by Firefox's task manager). This has led to system crashes for me if I have too many emanote tabs open and don't have an eye on my system's overall free memory.

**To Reproduce**
Run the emanote server using `emanote run` and then open the tab in Firefox.

**Expected behavior**
Not that big a memory footprint.

**Desktop (please complete the following information):**
Firefox 108.0.2.

**Additional context**
My Zettelkasten has quite a lot of math formulae (some of which are really long `align` environments with 50 and more lines), maybe this has something to do with this? However, the Zettels I have open right now only have 20ish math formulae each of which is rather short (only one to two handful of mathematical symbols) and they use 1GB RAM each nevertheless.

srid:
Could be something to do with Ema's websocket reload logic (which uses `morphdom`). See https://ema.srid.ca/topics/hot-reload

Here's the JavaScript code if anyone wants to help with this:
https://github.com/srid/ema/blob/master/ema/www/ema-shim.js

EDIT: Probably not; see my next comment below.

dpaetzel:
Just wanted to add: As a workaround I try to regularly close and reopen Emanote browser tabs in order to not lose my entire user session (there's this annoying systemd 252 behaviour where the entire X server gets reaped if one of the subprocesses uses too much RAM).

I did this just now (i.e. close an Emanote tab) and apparently *closing this one tab freed over 8GB of RAM* (that tab was open for several hours, though). :speak_no_evil: 

Sadly, I personally don't have the time right now to help with this.

srid:
I debugged this using Chrome inspector's memory profiling, and two things cause the memory leak:

1. highlightjs
2. stork re-registering

If you use a pre-stork Emanote (`nix run github:srid/emanote/0828297369dc4753134ff67287d0c303111f5c48`) with `highlightjs` removed from your `index.yaml` and Markdown frontmatter (you could switch to [`prism`](https://emanote.srid.ca/tips/js/syntax-highlighting#prism) which doesn't leak memory), the memory leak should go away. 

Obviously, we should fix both these issues. The stork leak is triggered by this re-registering:

https://github.com/srid/emanote/blob/a041483e055cb577db4bfca43bb4d528bd3626d3/default/templates/components/stork/stork-search-head.tpl#L55-L56

[...]

PR #672, opened 2026-04-26
Stork: drop inline IIFE for an ES module under _emanote-static/js/

srid:
**The Stork search controller was the last interactive behavior still living as an inline `<script>` IIFE in a Heist template.** This is the follow-up named in #663's body — Stork was deliberately deferred there because its `onclick="window.emanote.stork.toggleSearch()"` call sites span four templates and the migration warranted a focused change. Now it takes the same shape as the other behaviors landed in #663: a module under `_emanote-static/js/`, picked up by the importmap, no `window.emanote.stork` global, no marker `<script id="emanote-stork">` element.

Templates expose the search trigger via `data-emanote-stork-toggle`; a single delegated click listener on `document` covers the four sites (sidebar + breadcrumbs + minimal-layout header + modal backdrop) — the backdrop's role is "close" rather than "open", which works because the backdrop is only visible while the modal is open. The previously-separate inline theme-mirror script in `stork-search.tpl` (a `MutationObserver` re-skinning the search dialog when the user flips the dark-mode toggle) folded into the module too, since `document.documentElement` is never replaced by morph and the wrapper element re-resolves on each fire. Base URL comes from `document.baseURI` instead of a custom data attribute — same value, no marker element to maintain.

**A morph-time regression surfaced during review and is fixed in-PR**: the dialog rendered unstyled after a route switch because the new `#stork-wrapper` element gets no `stork-wrapper-edible{,-dark}` class — the `MutationObserver` only fires when `<html>.class` actually changes, and a plain morph nav doesn't toggle the theme. _The fix folds `applyStorkTheme()` into the existing `onMorph` callback that already marks the index stale_, so both post-morph concerns hang off the same hook. A new `@morph`-tagged scenario navigates-then-searches and asserts the wrapper carries one of the edible classes; it failed on the original code, passes now. Issue #673 tracks the broader auto-morph e2e mode that would surface this entire class of regression automatically across every behavior.

> **Stays inline by necessity.** The vendor `<script src=".../stork/stork.js">` (the WASM loader that defines `window.stork`) keeps loading from `<head>` as a non-module blocking script — it must parse before our module evaluates so `stork.initialize()` and `stork.register()` resolve. A module-load guard fails fast with a diagnostic if the load order ever breaks.
>
> **WASM init is once-per-session.** Same memory-leak trade-off as before (#411): `onMorph(markIndexAsStale)` flips a flag on each in-app navigation, and `refreshIndex` re-registers lazily the next time the user opens search.

E2e: four scenarios cover Ctrl+K opens, Esc closes, the sidebar search button opens, and the morph-then-search styling assertion. The `Then("the Stork search modal is {string}", …)` assertion checks the body's `stork-overflow-hidden-important` class (the actual open-state truth, not computed visibility — the container is `position:fixed` with a backdrop, so `getComputedStyle` reads `display: block` either way; only the class flips).

Eleven commits: the primary migration plus four hickey/lowy fixes (`searchShown` mirror flag → DOM-derived getter; vendor `window.stork` guard; comment correctness in `base.tpl`; morph-survival docstring allows combining strategies), three police fixes (silent URL-parse warn in `getBaseUrl`, `MODAL_HIDDEN_CLASS` constant, `registerIndex(forceOverwrite)` over options bag), an e2e step fixup, the morph-regression test + fix, and a merge from master. With this in, the inline-IIFE inventory named in #643 is fully closed.

### Try it locally

```sh
nix run github:srid/emanote/stork-es-module
```


srid:

| # | Lens   | Finding                                                              | Disposition      |
|---|--------|----------------------------------------------------------------------|------------------|
| 1 | Hickey | `searchShown` mirror flag could drift from body class (silent Esc)   | Fixed in this PR |
| 2 | Hickey | `window.stork` not guarded → cryptic error far from load-order cause | Fixed in this PR |
| 3 | Lowy   | `base.tpl` comment misnamed inline as "Stork controller"             | Fixed in this PR |
| 4 | Lowy   | `morph.js` docstring said "must choose ONE strategy" — Stork combines| Fixed in this PR |
| – | Both   | 9 No-op findings (intentional design, structurally sound)            | No-op            |

### Hickey rationale

(1) `searchShown` shadowed the `stork-overflow-hidden-important` body class. If anything outside the module ever mutated that class (a future utility, devtools), the keydown handler would silently no-op on Esc when the modal looked open. Replaced with `isSearchShown()` that queries the body class — one extra `classList.contains` per Esc keypress, eliminates the drift hazard.

(2) `window.stork` is a vendor global the module assumes is defined at evaluate time; the load-order constraint was documented in a comment but not enforced. Without the guard, a future template tweak that reorders or removes the vendor `<script src>` in `stork-search-head.tpl` would surface as `stork is not defined` on the user's first click — far from the actual cause. Added a `typeof stork === 'undefined' → throw` at module load with a pointer to the failing template.

### Lowy rationale

(3) `base.tpl`'s `<emanoteJsBundle />` comment described the still-inline Stork script as the "Stork controller", but Stork's controller is now an ES module — the inline is just the vendor WASM loader. Renamed to "vendor Stork WASM loader" so a future reader doesn't think the controller is still inline.

(4) The `morph.js` morph-survival docstring (added in #663 review) said "a behavior must choose ONE strategy". Stork now combines strategies 2 (event delegation), 3 (MutationObserver), and 4 (`onMorph`) — distinct concerns, distinct dependency shapes, all in one module. Reworded to allow combination explicitly with stork as the concrete example, so the next behavior author isn't misled.

### No-op findings (kept verbatim from sub-agent review)

- `data-emanote-stork-toggle` dual role (open + close) is fine — backdrop is only visible while modal is open, so `toggleSearch()` derives action from current state.
- `MutationObserver` filter scope is pragmatic — narrowing further to ".dark substring change" would add regex/parse overhead per mutation.
- `getBaseUrl()` via `document.baseURI` is a justified simplification, not a leaky abstraction (`<base href>` is unconditional in `base.tpl`).
- `registerIndex(forceOverwrite=…)` lazy-on-search trade-off (vs. eager-on-morph) is honest given typical browsing patterns.
- Stork stays cleanly separate from the four #663 behaviors (single-direction `import { onMorph } from '@emanote/morph'`).
- e2e step shapes match patterns elsewhere in the suite.

### `/code-police` follow-ups (3 commits)

- **Fact-check**: `getBaseUrl` URL-parse `catch` was silent → now logs `console.warn` so a malformed `document.baseURI` surfaces instead of search degrading to root with no signal.
- **Elegance**: `'stork-overflow-hidden-important'` literal hit the rule-of-three across `toggleSearch` / `clearSearch` / `isSearchShown` → extracted to `MODAL_HIDDEN_CLASS`.
- **Elegance**: `registerIndex(options)` had a one-field options bag; only one of two callers passed anything → replaced with explicit `forceOverwrite = false` boolean.

E2e: 21 scenarios; live=21/21, static=20+1-skipped (the `@morph` toc-spy scenario).

srid:

| sync | ✓ | 0s | forge=github; lost-crash worktree |
| research | ✓ | 1m 24s | 4 onclick call sites + inline theme-mirror + vendor `<script src>` (stays) |
| branch | ✓ | 6s | stork-es-module from origin/master |
| implement | ✓ | 3m 17s | stork.js (~125 LOC) + 4 attr swaps + theme-mirror folded in + 3 e2e scenarios |
| check | ✓ | 40s | cabal build all |
| docs | ✓ | 0s | CHANGELOG sub-bullet under existing #643 entry |
| fmt | ✓ | 16s | pre-commit clean |
| commit | ✓ | 33s | primary feature commit |
| hickey+lowy | ✓ | 6m 0s | 4 findings, 4 fixed; 9 No-ops |
| police | ✓ | 4m 0s | rules clean; fact-check 1 fix; elegance 2 fixes |
| test | ✓ | 5m 11s | e2e live 21/21; static 20+1-skipped |
| create-pr | ✓ | 1m 21s | draft PR + hickey/lowy comment |
| ci | ✓ | 1m 23s | `vira ci` passed on `7fac1359` |
| **Total** | | **24m 29s** | |

**Slowest step**: `hickey+lowy` (6m 0s) — two parallel sub-agent reviews dominated, similar shape to #663's resumed run.

### Optimization suggestions

- **`test` ran 5m 11s** — both modes had to rebuild after the dirty-tree warning interrupted Nix's cached output. A subsequent run of just `e2e-static` + `e2e-live` after the first build cache was warm took ~7s combined; the cold first-run cost is the real story.
- **`hickey+lowy` and `police` reviewed a much smaller diff than #663** (10 files / +213 −108 vs ~900 LOC) but ran for 10m combined. The cost is the agent latency itself, not anything about the diff size — running them in parallel inside a single response (which I did) is the only knob.
- **No CI retries** — five rounds of fixes (4 hickey/lowy + 3 police) all landed clean and `vira ci` passed first try.

Workflow completed at 2026-04-26T17:20Z. PR is ready for review: ships the last interactive behavior into the ES-module pattern established by #663, closes the inline-IIFE inventory named in #643.

PR #677, opened 2026-04-27
Pandoc tables: honour colSpec, cell metadata, and footers

### LINKED ISSUE #27: Pandoc: complete table formatting
There is a TODO here that should be done

https://github.com/srid/emanote/blob/e1a8e59acfd361da403aedfb695dc6d6117ddbcd/src/Heist/Extra/Splices/Pandoc.hs#L136-L138

Ref: https://github.com/jgm/commonmark-hs/blob/4f7d8b7b37dcc672d46f6bf967afbd47c589e050/commonmark-extensions/test/pipe_tables.md

commonmark-hs doesn't do captions, so we can just apply what's available - like colspecs, etc.

srid:
**Markdown tables now render the alignment, widths, spans, attrs, and footer rows that Pandoc actually parses** — until now the renderer was throwing all of that away and emitting only "rows of cells." The TODO that issue [#27](https://github.com/srid/emanote/issues/27) called out has been live since 2020. Closes #27.

The actual fix lives upstream in [`srid/heist-extra#15`](https://github.com/srid/heist-extra/pull/15) — emanote's side of the change is just the `flake.nix` pin bump, a one-table alignment demo in the user docs, and a CHANGELOG entry. Captions are *still* skipped: commonmark-hs doesn't emit them, and the existing TODO covers them separately.

The heist-extra branch went through two parallel hickey/lowy passes plus `/simplify`. The first hickey pass caught a real bug in the initial implementation — `zipWithM (renderCell tag) [0..]` conflated cell list-position with column index, silently misaligning every cell after a `ColSpan > 1`; fixed by `cellColumnIndices`, with a regression test. The second pass caught an older `concatAttr` defect surfacing under the new code path: a cell carrying its own `style` (from raw HTML in the source) plus the renderer's `text-align: …` produced two `style` attrs on a single `<td>`. *Fixed by a localized `mergeStyleKVs` helper at the cell-render call site* — `concatAttr` itself stays untouched so other renderers keep their existing semantics.

Coverage in heist-extra: five unit suites for the pure helpers (`alignmentStyle`, `colSpecsToColgroup`, `cellSpanAttrs`, `cellColumnIndices`, `mergeStyleKVs`) plus eleven integration scenarios that drive `rpBlock` end-to-end through `evalHeistT` and assert against the rendered HTML — every variant the new code path is supposed to honour, including the colspan-misalignment regression and the duplicate-style cell.

> **Merge order**: srid/heist-extra#15 should land first; this PR's pin then resolves to a tag on `master` instead of an unmerged feature branch.

### Try it locally

```sh
nix run github:srid/emanote/pandoc-table-formatting -- -L docs run
```

Then open `/guide/markdown` and scroll to the *Tables* section — the new alignment-demo table shows left/center/right column alignment honoured per cell.


srid:

The substantive code lives in [`srid/heist-extra#15`](https://github.com/srid/heist-extra/pull/15); two parallel hickey/lowy passes ran on that branch end-to-end, plus a `/simplify` pass. Two real defects were caught and fixed before this PR's pin landed.

| # | Lens   | Finding                                                          | Disposition       |
|---|--------|------------------------------------------------------------------|-------------------|
| 1 | Hickey | colSpan-aware column index (`cellColumnIndices`)                 | Fixed in heist-extra (commit 6022bb0) |
| 2 | Hickey | duplicate `style` attrs from `concatAttr` chain (`mergeStyleKVs`)| Fixed in heist-extra (commit f819fd2) |
| 3 | Hickey | `let` vs `where` asymmetry in `rpBlock'`                         | No-op (forced by closure capture) |
| 4 | Hickey | four exported helpers — factoring                                | No-op (each is one orthogonal axis) |
| 5 | Hickey | `B.Table` arm decomplecting opportunity post-simplify            | No-op (right granularity) |
| 6 | Lowy   | alignment vs col-width axes                                      | No-op (correctly split) |
| 7 | Lowy   | tailwind injection boundary still incomplete (`TODO` preserved)  | Deferred — explicitly scoped out of #27 by the user; tracked by the existing inline `TODO` |
| 8 | Lowy   | tables lack `withTplTag` hooks like Para/Header                  | Deferred — new feature, not a fix to #27; coupled to the same TODO |
| 9 | Lowy   | `evalHeistT` test setup vs `emptyTemplateState`                  | No-op (correct choice — pure helpers don't consult `HeistState`) |

### Hickey rationale

The first pass caught the structural bug in the initial implementation: `zipWithM (renderCell tag) [0..]` conflated cell *list position* with *column index*, silently misaligning every cell after a `ColSpan > 1`. Fixed with `cellColumnIndices` (a single-pass walk that accumulates the running column offset) and pinned by an integration test that builds a 4-column table with a span-2 cell and asserts the right cell gets the right alignment.

The second pass caught the still-live `concatAttr` style-key duplication: when a cell already carries its own `style` (e.g. from raw HTML in the source markdown) and the renderer adds a `text-align: …`, the `attr1 <> attr2` list-append produced two `style` attributes on a single `<td>` — invalid HTML, with browsers silently dropping one of the values. *Fixed locally* via a `mergeStyleKVs` helper at the cell-render call site; `concatAttr` itself stays untouched so other renderers keep their existing semantics. Pinned by both a unit suite and an integration test that builds a cell with `style="color: red"`, renders it under a right-aligned column, and asserts a single `style="color: red; text-align: right"` lands on the output.

The other Hickey findings — closure capture in the `B.Table` arm and the `let` vs `where` asymmetry — are structurally forced (the local renderers close over four bindings that no other block arm uses). No-op.

### Lowy rationale

Volatility-encapsulation lens: the four pure helpers each capture one orthogonal axis of change (alignment-vocabulary mapping, width-spec formatting, span-attr emission, column-index resolution under spans), and the split is correct — alignment and col-widths are genuinely different volatility types. The Tailwind classes hard-coded inside the `B.Table` arm represent a real injection-boundary gap, but the user explicitly scoped tailwind cleanup out of this PR (the existing `-- TODO: Move tailwind styles to pandoc.tpl` is preserved unchanged). Tables also lack `withTplTag` template-override hooks that Para/Header have — a real feature gap, but not a fix to issue #27; both deferrals are tracked alongside the same TODO since they all depend on the eventual template migration.

srid:
## Evidence

PR #677 teaches Emanote's table renderer to honour Pandoc's column alignments (and the related cell metadata that Pandoc tracks). On `master` the renderer dropped that information and emitted `<th>`/`<td>` with no `text-align` style, so Markdown headers like `| :--- | :---: | ---: |` were silently ignored. After this change the renderer threads alignment through to inline `style='text-align: …'` on every cell. Both screenshots below render the same `Tables` section of `docs/guide/markdown.md` against the same notebook — only the binary differs.

Before — alignment metadata discarded; "Apple/Banana/Pear" and "Carrot/Eggplant/Kiwi" all hug the left edge despite the `:---: ---:` markers in the header.
*Master binary: every cell defaults to left-aligned regardless of the header markers.*

After — Pandoc alignment honoured: column 1 left, column 2 centered, column 3 right; headers shift to match.
*PR #677 binary: Apple/Carrot stay left, Banana/Eggplant center, Pear/Kiwi right-align — exactly what `| :----- | :------: | ----: |` requests.*

srid:

| research | ✓ | 6m 54s | Issue #27 read; TODO at heist-extra Render.hs identified; plan formed |
| branch | ✓ | 12s | Created emanote/pandoc-table-formatting and heist-extra/complete-table-rendering |
| implement | ✓ | 6m 47s | First heist-extra commit pushed; emanote pin bumped; alignment demo added; helpers + hspec coverage green |
| check | ✓ | 1m 28s | `cabal build all` clean against new pin |
| docs | ✓ | 18s | CHANGELOG Unreleased + alignment demo in markdown.md |
| fmt | ✓ | 13s | pre-commit (cabal-fmt, fourmolu, hlint, nixpkgs-fmt) all green |
| commit | ✓ | 23s | Primary feature commit pushed |
| **hickey+lowy** | ✓ | **39m 26s** | Two parallel passes on heist-extra. Caught 2 real defects (colspan/index conflation, duplicate style attrs); both fixed with regression tests. 2 lowy defers user-scoped. 2 simplify-pass refactors. PR comments posted. |
| police | — | 13s | Skipped: emanote PR diff is docs + flake-pin only |
| test | ✓ | 23s | 43 unit tests pass |
| create-pr | ✓ | 1m 15s | Draft PR opened |
| ci | ✓ | 7m 41s | vira ci ✅ on both repos; e2e-live + e2e-static + e2e-morph all green |
| evidence | ✓ | 9m 56s | Before/after screenshots posted |
| **Total** | | **1h 15m 26s** | |

**Slowest step**: `hickey+lowy` (39m 26s — dominated by two parallel sub-agent passes plus follow-up commits with regression tests).

### Optimization suggestions

- **`hickey+lowy` ate 52% of the run.** The first-pass review on the initial implementation was unavoidable, but the *second* pass (after `/simplify`) only ran because the user explicitly asked for it. For routine /do invocations, a single hickey+lowy pass on the post-simplify diff would skip ~half the duration.
- **Dual-repo CI re-runs after the `Render.Internal` split cost ~7m.** Each pin re-bump on the emanote side triggers a fresh `vira ci` against the new HEAD. Batching the heist-extra refactors more (one commit instead of two simplify commits + one Internal-split commit) would halve the number of pin bumps.
- **`evidence` took 10m** because both binaries (master and the PR HEAD) are built from scratch via Nix. Cache hits would cut this dramatically; if the user runs `/do` against `master`-adjacent PRs frequently, keeping a warm `evidence-assets`-style notebook image around would help.
- **`research` took 7m** because the AST types (`pandoc-types` `ColSpec`/`Cell`/`Alignment`) had to be cloned and read from scratch. For Pandoc-related work, a one-line note in `.agency/do.md` pointing to `/tmp/pandoc-types-src` (or a similar known-good clone) would skip the discovery dance.

Workflow completed at 2026-04-26T20:43Z.

PR #682, opened 2026-04-27
Stop splitting URLs out of Markdown link labels (#349)

### LINKED ISSUE #349: URIs contained in link description shouldn't be converted to separate hyperlinks
**Describe the bug**

The following markdown:

```
[https://www.website.com](https://www.website.com#something)

[nobody@example.com](mailto:nobody@example.com?subject=Some%20subject)

[A website similar to https://www.foo.com and https://www.bar.com](https://www.baz.com)
```

is converted to the following HTML (at least in live mode):

```html
<p class="mb-3">
	<a href="https://www.website.com#something" class="text-blue-600 hover:underline" target="_blank" rel="noopener"></a>
	<a href="https://www.website.com" class="text-blue-600 hover:underline" target="_blank" rel="noopener">https://www.website.com</a>
</p>
<p class="mb-3">
	<a href="mailto:nobody@example.com?subject=Some%20subject" class="text-blue-600 mavenLinkBold hover:underline"></a>
	<a href="mailto:nobody@example.com" class="text-blue-600 mavenLinkBold hover:underline">nobody@example.com</a>
</p>
<p class="mb-3">
	<a href="https://www.baz.com" class="text-blue-600 hover:underline" target="_blank" rel="noopener">A website similar to </a>
	<a href="https://www.foo.com" class="text-blue-600 hover:underline" target="_blank" rel="noopener">https://www.foo.com</a>
	and
	<a href="https://www.bar.com" class="text-blue-600 hover:underline" target="_blank" rel="noopener">https://www.bar.com</a>
</p>
```

**Expected behavior**
I'd expect the following result:

```html
<p class="mb-3">
	<a href="https://www.website.com#something" class="text-blue-600 hover:underline" target="_blank" rel="noopener">https://www.website.com</a>
</p>
<p class="mb-3">
	<a href="mailto:nobody@example.com?subject=Some%20subject" class="text-blue-600 mavenLinkBold hover:underline">nobody@example.com</a>
</p>
<p class="mb-3">
	<a href="https://www.baz.com" class="text-blue-600 hover:underline" target="_blank" rel="noopener">A website similar to https://www.foo.com and https://www.bar.com</a>
</p>
```

Arguably in the third example the current behavior makes some sense. But definitely not in the first two.

srid:
**A Markdown link whose label contains a URL or email address now renders as a single hyperlink**, not as the parent link plus separate autolinks for each embedded URL. `commonmark-hs`'s autolink extension greedily turns the inner URL into its own `Link` even inside a parent link's label — including positions wrapped in emphasis (`Strong`, `Emph`, `Span`, …) — and Pandoc renderers then split the result into multiple `<a>` tags. That's the visible bug from [#349](https://github.com/srid/emanote/issues/349).

The fix is a new `flattenNestedLinks` pass added as the **last step of `Emanote.Pandoc.BuiltinFilters.preparePandoc`**, sitting after `linkifyInlineTags`, `fixEmojiFontFamily`, and `setExternalLinkIcon`. It walks every `Link`'s label, drops any inner `Link` wrapper at any depth, and keeps the inner inlines intact so the displayed label survives. _Image children and non-link container inlines are preserved structurally; only `Link` wrappers vanish._

A new `Emanote.Pandoc.BuiltinFiltersSpec` pins the function's behaviour through the public `preparePandoc` entry point for the URL-only, mailto, sentence-with-multiple-URLs, and emphasis-wrapped cases from the bug report. The spec reuses the existing `plainify` helper from `commonmark-wikilink` rather than hand-rolling a Pandoc plain-text walker.

End-to-end coverage: a new fixture note `tests/fixtures/notebook/issue-349.md` exercises all four cases, and `tests/features/smoke.feature` adds a scenario that asserts (a) each parent link's `href` + label resolves correctly and (b) **exactly four `<a>`s** target the case URLs in the article body. The count assertion is the strongest regression guard — every bug mode adds spurious anchors.

> Earlier revisions of this PR routed the fix through `commonmark-simple` and through a post-user-filter wrapper in `Note.hs`. Both placements turned out to be defending against hypothetical risks rather than the reported bug. Folding the pass into `preparePandoc` is the principled call: user filters get the last word in emanote's pipeline by design, and the only confirmed bug is the parser artifact, which `preparePandoc` fully addresses.

### Try it locally

```sh
nix run github:srid/emanote/fix/349-flatten-nested-links
```

Closes #349.


srid:

| # | Lens  | Finding                                                  | Disposition      |
|---|-------|----------------------------------------------------------|------------------|
| 1 | Lowy  | `flattenNestedLinks` should not be exported              | Fixed in this PR |

Hickey returned **no findings**.

### Hickey rationale

The evaluation traversed all seven layers and surfaced nothing actionable:

- **`flattenNestedLinks` itself.** No complecting. The function delegates tree traversal entirely to `Text.Pandoc.Walk` — it does not implement traversal. Within the one pattern-match case, it does exactly one thing: replace the children of a parent `Link` with the result of unwrapping inner `Link`s at any depth in its subtree. The three operations the question framed separately (find, rewrite, drop) are logically inseparable here — they are the single action of "normalize a node." There is no alternative decomposition that would reduce conceptual load.
- **Wiring at both `parseMarkdown` and `parseMarkdownWithFrontMatter`.** Repetition, not duplication. Both call sites apply the same function with no variation, at symmetric positions (immediately before `return`). The rule is mechanical — _"apply normalization to every Pandoc value we produce"_ — and the two sites are the only two output boundaries. A single common helper would require either a third function wrapping both, or restructuring the parsing pipeline. Current form is clearer: every production path in the module visibly terminates with the same normalization.
- **Test-side `links` / `plain` walkers.** Appropriate boundary isolation. The Emanote spec is testing the integration boundary (does the parser return the normalized form?), not the `flattenNestedLinks` function itself. The local walkers extract facts from the Pandoc AST without importing anything from commonmark-simple's test suite. If commonmark-simple's test helpers changed or were removed, the Emanote spec would be unaffected — the right property for a dependency boundary.
- **`common shared` cabal stanza.** Earning its keep. The stanza consolidates `build-depends`, `default-language`, and `ghc-options` shared between the library and the new test suite. Without it, those fields would be repeated verbatim and a future GHC options change would require updating both. Standard Cabal pattern.

### Lowy rationale

**Volatility named.** The autolink extension in `commonmark-hs` produces nested `Link` nodes when a URL or email appears inside a link label. This is a _parser artifact_ — an impedance mismatch between commonmark-hs's AST shape and what downstream consumers expect. The volatility axis is _how commonmark-hs's extension pipeline shapes its output_.

**Boundary verdict — fix is at the right boundary.** `commonmark-simple`'s stated contract is to provide a stable, normalized interface over `commonmark-hs`. An autolink artifact is a `commonmark-hs` concern, not an Emanote concern. Placing the fix in Emanote would be functional decomposition (Emanote owns parsing) rather than volatility-based (the volatility belongs to the upstream library layer). Every future `commonmark-simple` consumer gets the fix automatically; no re-implementation drift. The alternative — a PR to `commonmark-hs` itself — is ideal long-term but not blocking.

**Silent normalization.** Automatic, unconditional normalization inside `parseMarkdown` and `parseMarkdownWithFrontMatter` is correct. The fix converts broken output into correct output; it is not a feature toggle. Opt-in would mean callers must know about and handle the artifact themselves — the definition of a leaky boundary. The interface signature (`Pandoc` in, `Pandoc` out) does not change. The normalization is a no-op if commonmark-hs ever stops producing nested links.

**Test duplication.** The Emanote regression spec pins the contract of its dependency at the parser boundary — testing that `commonmark-simple` delivers what it promises, not how `flattenNestedLinks` works internally. The four cases are the minimal regression set. Control cases (normal link, bare autolink) live in `commonmark-simple` and need not be echoed. This is layered defense, not maintenance liability.

**Finding 1 (applied).** `flattenNestedLinks` had been added to the public export list, but it has zero production consumers — neither Emanote nor the `commonmark-simple` tests call it directly. Exporting it would have signalled that the boundary is permeable: callers could call it separately, bypassing the normalization-at-parse-time contract. It is implementation detail of the normalization pipeline. Dropped from the export list (`commonmark-simple` commit `284bf71`); the function itself stays for internal use and can be re-exported if a real consumer turns up.

srid:
## Evidence

A four-case test notebook (URL-only label, mailto label, sentence containing embedded URLs, URL inside emphasis) reveals the regression and confirms the fix.

**Before** (master) — each link splits into a stray empty `<a>` plus a separate autolink, and the sentence case fragments into three hyperlinks:

Before — master renders broken split links

**After** (PR branch) — every case renders as exactly one anchor pointing at the author-supplied URL:

After — PR #682 renders one link per case

PR #682 closes #349 by flattening the inner `Link` at the parser boundary so a single hyperlink renders as written.

srid:

| research | ✓ | 4m 53s | issue #349 understood; planned fix in commonmark-simple per maintainer; located parseMarkdown at `emanote/src/Emanote/Pandoc/Markdown/Parser.hs:15`, sole non-test caller at `Note.hs:421`; `flake.nix:30` wires commonmark-simple input |
| branch | ✓ | 14s | on branch `fix/349-flatten-nested-links` (renamed from worktree branch); HEAD == `origin/master` |
| implement | ✓ | 7m 26s | implemented `flattenNestedLinks` in commonmark-simple; wired emanote flake to the branch; added regression spec + CHANGELOG entry |
| check | ✓ | 1m 33s | `cabal build all` succeeded; lib + emanote exe + test-suite linked |
| docs | ✓ | 11s | CHANGELOG Unreleased entry added; README/docs unaffected (parser-level bug fix) |
| fmt | ✓ | 38s | `just fmt`: cabal-fmt/fourmolu/hlint/nixpkgs-fmt all passed |
| commit | ✓ | 1m 38s | primary commit `fc47b150`; commonmark-simple branch pushed |
| hickey+lowy | ✓ | 5m 38s | hickey: no findings; lowy: 1 finding (`flattenNestedLinks` unexported) — fixed in `284bf71` (dep) + `ebe6350b` (emanote) |
| **police** | ✓ | **23m 5s** | rules clean; fact-check fixed deeper `Strong`/`Emph` nesting (`053f6b4` dep + `32d7d3b3` emanote); elegance reused `plainify` from commonmark-wikilink (`02e96f96`) |
| test | ✓ | 23s | `cabal test all`: 47 examples, 0 failures (incl. 4 new #349 cases) |
| create-pr | ✓ | 2m 31s | draft PR #682 + dep PR srid/commonmark-simple#2 created; hickey/lowy analysis posted |
| ci | ✓ | 2m 10s | `vira ci` + `just e2e-{live,static,morph}` all exit 0 (parallel); HEAD `acec6b04` unchanged |
| evidence | ✓ | 2m 39s | before/after screenshots posted on a four-case test notebook |
| **Total** | | **55m 5s** | |

**Slowest step**: `police` (23m 5s).

### Optimization suggestions

- **`police` dominated at 23m (42% of wall time)**, driven by the fact-check sub-pass spotting the deeper `Strong`/`Emph` nesting case — which produced one extra dep round-trip (commit + push + flake-lock bump + emanote rebuild) and the elegance pass reusing `plainify`. Both findings were real, but for similar-shape parser bugs, **enumerate every container constructor up-front** (`Strong`, `Emph`, `Span`, `Underline`, …) before writing the first walker; it would have collapsed the two trips into one.
- **Two `commonmark-simple` redirects via the flake input** burned ~3m total. Once the dep PR opens, **prefer a single rev pin** (`?rev=<sha>`) at the start of the run rather than a branch ref that gets bumped multiple times — each bump rebuilds the local Haskell tree.
- **`research` at 4m 53s** spent on `Explore` discovery that overlapped with `git log -p` reads in the main context; for a tightly-scoped parser bug like this, skip the broad codebase walk and read `Parser.hs` + `BuiltinFilters.hs` + the cabal test-suite stanza directly.
- **`hickey+lowy` at 5m 38s in parallel sub-agents** is fine — no easy speedup unless the prompts are pre-canned for "single-function dep + thin emanote bump" shape.

Workflow completed at 2026-04-27T13:05:52Z.

srid:
## Architectural pivot — single source of truth in emanote

The earlier revision placed `flattenNestedLinks` in `commonmark-simple` (PR [#2](https://github.com/srid/commonmark-simple/pull/2)) and applied it inside `parseMarkdown`. A second [Hickey/Lowy](https://kolu.dev/blog/hickey-lowy/) pass surfaced a bypass that the first pass missed: `Emanote.Model.Note.parseNoteMarkdown` runs user-defined Pandoc filters *after* `parseMarkdown` returns. A user filter producing `Link`-in-`Link` would slip past a parse-time normalization and reach the renderer as invalid nested `<a>` HTML.

The Lowy *"every `commonmark-simple` consumer benefits automatically"* argument is real but applies only to consumers without a post-parse user-filter step (`imako`, `actualism-app`, `ka` — none have reported the bug). For emanote specifically, parse-time application is *insufficient* because of user filters, and `commonmark-simple` placement gives a false sense of security at the parse boundary.

Folded the fix into emanote at the render boundary (`Note.hs:443`):

```haskell
doc <- flattenNestedLinks <$> applyPandocFilters scriptingEngine filterPaths (preparePandoc doc')
```

Single placement, single source of truth, applied as the postcondition of the *whole* pre-render pipeline. The dep PR is closed; `flake.nix` reverts to `commonmark-simple` `0.2.0.0`. If a non-emanote consumer ever hits the same bug, the function can be promoted to `commonmark-simple` then.

### Updated [Hickey/Lowy](https://kolu.dev/blog/hickey-lowy/) ledger

| # | Lens   | Finding                                                           | Disposition       |
|---|--------|-------------------------------------------------------------------|-------------------|
| 1 | Hickey | Parse-time fix bypassable by user filters → render-boundary fix   | Fixed in this PR  |
| 2 | Lowy   | Original commonmark-simple placement encapsulated only the parser-quirk axis; emanote needs the filter-pipeline-output axis too | Fixed in this PR  |
| 3 | Hickey | Mild module-purpose drift in `BuiltinFilters` (post-filter postcondition lives next to pre-filter passes) | No-op — call-site comment in `Note.hs:443` carries intent; one boundary crossing beats a new module for a single function (Lowy concur) |

### Verification

- `cabal test all`: 47 examples, 0 failures (`Emanote.Pandoc.BuiltinFiltersSpec` covers all 4 #349 cases).
- `vira ci` + `just e2e-{live,static,morph}`: all exit 0 against HEAD `06fe6e00`.

srid:
## Architectural simplification — `flattenNestedLinks` is just a `preparePandoc` pass

The previous revision applied `flattenNestedLinks` *after* user-defined Pandoc filters as a "guaranteed postcondition" against a hypothetical user filter producing `Link`-in-`Link`. That hypothetical isn't a reported bug, and silently flattening user-filter output overrides the design contract that user filters get the last word in emanote's pipeline.

Folded into `preparePandoc`'s `>>>` chain instead — the only confirmed bug (#349) is the `commonmark-hs` autolink parser artifact, which `preparePandoc` fully addresses. Single placement, single source of truth, and `Note.hs:443` is back to its pre-PR shape (`applyPandocFilters … (preparePandoc doc')`).

### E2E coverage added

`tests/features/smoke.feature` now carries a `(regression: #349)` scenario backed by a new fixture `tests/fixtures/notebook/issue-349.md`. The four per-case assertions confirm each parent link's `href` + label resolves correctly; the count assertion (**exactly 4 article anchors targeting `issue349` URLs**) is the strongest regression guard — every bug mode adds spurious autolink anchors, so the count goes up the moment someone breaks the pass.

### Verification (HEAD `ff330aac`)

- `cabal test all`: 47 examples, 0 failures.
- `vira ci` + `just e2e-{live,static,morph}`: all green.

## Evidence (refreshed for HEAD `ff330aac`)

Refreshed before/after screenshots for the four cases from #349. The before-state was captured against `master` using a standalone reproduction notebook; the after-state was captured against this branch using the new `tests/fixtures/notebook/issue-349.md` fixture, which is the same notebook the e2e scenario asserts on.

Before (master, broken — each row splits into a parent link plus a stray autolink for the embedded URL/email):

before

After (this branch, fixed — each row renders as exactly one anchor, label text preserved verbatim):

after

Closes #349.

PR #696, opened 2026-05-02
Percent-encode tag-index URLs so `##` tags work

### LINKED ISSUE #199: Special tags not properly linked
I use Zettelkasten like Zettelkasten.de, which they use the concept of structure notes, found by following the tag `##.....`.
Emanate seems to fetch this type of note, see below

Capture d’écran 2021-11-25 à 22 32 21

But when clicking to it (`##§1`) I get this link: `https://notes.danielmathiot.tech/-/tags/#§1`, but they are not fetched in the tag index (see here: https://notes.danielmathiot.tech/-/tags/#§1)

srid:
Tag links containing characters reserved by RFC 3986 — most importantly `#`, used in Zettelkasten "structure note" tags like `##§1`, and any non-ASCII glyph — currently produce a `href` the browser truncates at the literal `#` and treats as an HTML fragment. **Both call sites that emit a tag-index URL now route through a percent-encoder so the link survives the trip into `href`.** Closes #199.

### Where the bug lived

| Path | Bug | Fix |
| --- | --- | --- |
| `Emanote.Pandoc.BuiltinFilters.linkifyInlineTags` (rewrites inline `#tag` syntax) | `encodeRoute . encodeTagIndexR` produced raw text via `Slug.unSlug` | New `encodeTagIndexUrl :: [HT.TagNode] -> Text` maps `Slug.encodeSlug` over each path segment |
| `default/templates/components/metadata.tpl` (page-metadata tag chips) | Pasted raw `${value}` into the `href` attribute — the original FIXME comment called this out | New `ema:tagsList` Heist splice exposes per-tag `${ema:tag:url}` (computed via `siteRouteUrl`, which already routes through Ema's `filepathToUrl` encoder) and `<ema:tag:name />` |

The two encoding paths exist because the Pandoc filter has no `Model` reachable and so cannot call `siteRouteUrl`. _The new `encodeTagIndexUrl`'s Haddock spells out the invariant — both encoders must produce the same output under `UrlDirect`._

### Regression coverage

Four pinned tests in `BuiltinFiltersSpec`:

- `tagged #foo` → `-/tags/foo.html`
- `tagged #foo/bar` → `-/tags/foo/bar.html`
- `tagged ###structure` → `-/tags/%23%23structure.html`
- `tagged #§1` → `-/tags/%C2%A71.html`

### Try it locally

```sh
nix run github:srid/emanote/fix-tag-url-encoding
```


srid:

| # | Lens   | Finding                                                                  | Disposition       |
|---|--------|--------------------------------------------------------------------------|-------------------|
| 1 | Hickey | Two independent encoding paths for tag-index URLs share no code          | Fixed in this PR  |
| 2 | Hickey | `encodeTagIndexUrl` had no note that it must agree with Ema's encoder    | Fixed in this PR  |
| 3 | Lowy   | Path B (metadata template) allegedly emits unencoded URLs                | No-op (incorrect) |
| 4 | Lowy   | `encodeTagIndexUrl` is a workaround, not a volatility encapsulation      | Fixed in this PR  |
| 5 | Lowy   | Template contract shift from JSON binding to typed Haskell splice        | Fixed in this PR  |

Findings 1, 2, and 4 collapsed into a single Haddock-clarifying commit (`14414599`) — the invariant between `encodeTagIndexUrl` and `siteRouteUrl` is now spelled out at the function definition. Finding 5 became a splice-contract docstring (`bff3768a`).

Lowy's Finding 3 was rejected after verifying via ghci: `Ema.routeUrlWith` _does_ percent-encode through `filepathToUrl` → `Slug.encodeSlug` (see `~/code/ema/ema/src/Ema/Route/Url.hs:48-50`), so the metadata-template path is correct as written. Output for the worked example: `Ema.routeUrlWith UrlDirect rp (encodeRoute (encodeTagIndexR [TagNode "##structure"]))` returns `"-/tags/%23%23structure.html"`.

### Hickey rationale

The diff introduces two independent code paths for the same domain concept ("the percent-encoded URL for a tag-index page given a list of `TagNode`s"):

- **Path A — Pandoc filter**: `encodeTagIndexUrl . toList . HT.deconstructTag` in `BuiltinFilters.hs`, manually mapping `Slug.encodeSlug` over the route segments.
- **Path B — Heist splice / TagIndex navigation**: `siteRouteUrl model (tagIndexRoute …)` in `Common.hs` and `View/TagIndex.hs`, which routes through Ema's `routeUrlWith`.

The split is real architecture, not laziness — the Pandoc filter has no `Model` in scope. But the coupling invariant (for identical input both paths must produce identical output) was previously enforceable only by code-review discipline. The follow-up Haddock on `encodeTagIndexUrl` makes the invariant load-bearing in the source: it states why the function exists separately, and that if `Slug.encodeSlug` or `encodeTagIndexR` changes, both encoders must be kept in lockstep.

`encodeTagIndexUrl` and `encodeTagIndexR` sitting side-by-side in `Type.hs` is a mild scope creep for the module (route-shape construction now coexists with URL-string encoding), but the structural problem is _not_ the module boundary — it's the absence of a mechanical relationship between the two encodings. The invariant Haddock is the cheapest move that makes the relationship visible without threading `Model` through the Pandoc filter (which would be a much wider refactor).

### Lowy rationale

The volatility being encapsulated is "tag-index URLs may contain RFC 3986 reserved characters and so must be percent-encoded before reaching `href`." That volatility is now encapsulated _twice_ — once by Ema's encoder (`filepathToUrl` / `Slug.encodeSlug`), and once by the new `encodeTagIndexUrl` helper. The Lowy test asks: does the interface change when the encapsulated volatility changes? Today, no — both encoders agree under `UrlDirect`. But the documentation needs to make the dual-encapsulation visible so a future maintainer doesn't change one without the other.

The metadata template's switch from a JSON-driven binding (`<with var="tags">`) to a typed Haskell splice (`<ema:tagsList>`) shifts the URL-computation locus from template into Haskell. That's correct (URL encoding genuinely lives in Haskell — Ema's encoder, the per-site URL strategy), but it changes the contract: template authors who want a precomputed URL now read it from `${ema:tag:url}` instead of pasting `-/tags/${value}`. The follow-up splice contract docstring spells out what `${ema:tag:url}` and `<ema:tag:name />` contain, including the percent-encoding and URL-strategy guarantees, so the new contract is discoverable from the splice definition.

The deeper Lowy question — should `Slug.unSlug` itself percent-encode in URL contexts, eliminating both `filepathToUrl`'s post-pass and `encodeTagIndexUrl` entirely? — crosses into the upstream `ema` and `url-slug` packages, so it's outside this PR's scope.

srid:
## Evidence

| Path | Before (master `d98c9ab`) | After (`01a1782a`) |
| --- | --- | --- |
| Page-metadata tag chips | before | after |
| Tag-index page after clicking inline `##§1` | before | after |

**Captured `href` values on this branch (commit `01a1782a`)** for a note frontmatter-tagged `["##§1", "##structure"]` with an inline `##§1`:

- Inline `##§1` link: `-/tags/%23%C2%A71.html`
- Metadata chip for `##§1`: `-/tags/%23%C2%A71.html`
- Metadata chip for `##structure`: `-/tags/%23%23structure.html`

Clicking the inline link navigates to `http://localhost/-/tags/%23%C2%A71.html`, whose page heading is `##§1 - Tag Index` — proving the URL survives intact.

**Master rendered the same elements as**:

- Inline `##§1` link: `-/tags/.html#§1.html` (the unescaped `##` collapses the path entirely; the rest becomes a fragment)
- Metadata chip for `##§1`: `-/tags/#§1.html`
- Metadata chip for `##structure`: `-/tags/##structure.html`

Clicking the inline link on master keeps the browser on the same page (URL becomes `…/-/tags/.html#§1.html`, a same-document fragment scroll), never reaching the tag-index page. The before/after rendered DOMs look visually identical because the difference lives in `href` attributes — the load-bearing evidence is the captured `href` strings above. Both inline (`Emanote.Pandoc.BuiltinFilters.linkifyInlineTags`) and metadata-chip (`default/templates/components/metadata.tpl`) call sites now percent-encode `#` and non-ASCII codepoints.

srid:

| research | ✓ | 9m 18s | Identified bug locations: `BuiltinFilters.hs:34-35` and `metadata.tpl:9`; planned `encodeTagIndexUrl` helper + `ema:tagsList` splice + tests |
| branch | ✓ | 5s | On feature branch `fix-tag-url-encoding` (off `origin/master`) |
| implement | ✓ | 3m 40s | Added `encodeTagIndexUrl` helper, switched `BuiltinFilters` to use it, added `ema:tagsList` splice, updated `metadata.tpl`, added 4 regression tests |
| check | ✓ | 43s | `cabal build all` succeeded (60/60 modules + emanote exe) |
| docs | ✓ | 38s | Added CHANGELOG entry under Unreleased Bug fixes |
| fmt | ✓ | 11s | fourmolu / hlint / cabal-fmt / nixpkgs-fmt all passed |
| commit | ✓ | 24s | Committed `dc90dd50` and pushed to `origin/fix-tag-url-encoding` |
| **hickey+lowy** | ✓ | **10m 5s** | Lowy Finding 1 (alleged metadata-template encoding bug) verified incorrect via ghci — Ema's `routeUrlWith` does percent-encode via `filepathToUrl`. Documentation findings addressed in 2 follow-up commits (`14414599` + `bff3768a`) |
| police | ✓ | 6m 7s | All 3 passes clean after 3 elegance commits (`1ddbadc5`, `0ab9aa00`, `01a1782a`) |
| test | ✓ | 36s | `cabal test all`: 66/66 examples passed including 4 new #199 regression tests |
| create-pr | ✓ | 1m 23s | Created draft PR #696; posted Hickey/Lowy analysis comment |
| ci | ✓ | 1m 49s | `vira ci` passed (signoffs at HEAD `01a1782a` for x86_64-linux + aarch64-darwin); e2e-live 31/31, e2e-static 29/31 (2 live-only skipped), e2e-morph 31/31 |
| evidence | ✓ | 3m 34s | Posted `## Evidence` comment with before/after screenshots + captured `href` values |
| **Total** | | **38m 44s** | |

**Slowest step**: `hickey+lowy` (10m 5s)

### Optimization suggestions

- **Pre-clone the Ema flake input** before `/do` for URL/slug-related fixes. `research` spent ~1-2m locating and reading `~/code/ema/ema/src/Ema/Route/Url.hs` and the `url-slug` package source — having both on disk in advance compresses that step.
- **Hickey+Lowy is the natural ceiling at ~10m** because the two agents run in parallel and the slower review dominates. Passing `--review-model=haiku` would cut it but trade depth; the structural finding here ("two encoding paths share no code, document the invariant") is the kind of catch worth the wait.
- **Evidence step rebuilt master from scratch** (`nix --refresh build github:srid/emanote/master`, ~1m). For repeat fixes against the same base, an out-of-band pre-build would be a free 1m.
- **Police elegance pass produced 3 separate fmt+commit+push cycles** (~10s each, ~30s total overhead). Acceptable for the per-finding commit-history clarity that `/do` enforces — flagged here for awareness, not as something to change.

Workflow completed at 2026-05-02 14:14 UTC.

Issue #352, opened 2022-09-24
Tags declared in `path.yaml` attach to file, but not to entire tags system


wyattjsmith1:
**Describe the bug**
Creating tags in a `.yaml` file creates the tag on the file, but the tag is not visible in the tags page, and clicking on the tag 404s.

**To Reproduce**
1. Create a file tree like this:
```
some_dir/
- my_file.md
some_dir.yaml
```
Where `some_dir.yaml` has the following:
```yaml
tags:
- my_tag
```
2. Generate and visit the `my_file.md` page.
3. Observe the tag `my_tag` exists on the page.
4. Click `my_tag`; observe a 404.
5. Go to the root tags page (click the 🏷️ icon at the bottom of every page)
6. Observe there is no entry for `my_tag`

**Expected behavior**
- Clicking tag should not result in 404.
- All declared tags should appear in the tags page

**Screenshots**
NA

**Desktop (please complete the following information):**
 - Browser: Firefox
 - Version: 105

**Additional context**
NA

PR #701, opened 2026-05-04
Union cascade arrays so child-note tags don't clobber folder tags

### LINKED ISSUE #697: Cascade tags clobbered when child note carries its own `tags`
## What

`Emanote.Model.SData.mergeAesons` uses `aeson-extra`'s `lodashMerge`,
which combines arrays element-by-element with `alignWith` (right wins
at each shared index). For the `tags` key this means: when a parent
YAML in the cascade declares `tags: [foo]` and a child note declares
its own `tags: [bar]` (or has any inline `#hashtag` in its body —
`addTagsFromMarkdown` writes those into `_noteMeta.tags`), the merge
emits `tags: [bar]`, not `tags: [foo, bar]`.

The cascaded tag silently disappears at the leaf, both in the
per-page chip-strip and (post-#352) in the global tag index.

## Why this is a problem

The cascade's job is "defaults that descendants can extend". For
scalar fields override-with-rightmost is correct semantics. For a
multi-valued field like `tags`, override-at-each-array-index is
neither replace-the-whole-list nor extend-the-list — it's a third
behavior driven by lodash's view of arrays as "objects keyed by
numeric indices". It's the source of long-standing confusion (see
e.g. the explicit comment in
`tests/fixtures/notebook/issue-352/note.md` warning future
contributors that the fixture must avoid inline `#tags` for the
cascade to survive).

## Repro

```
folder/
  note.md         # body: "intro #internal-note discussion"
folder.yaml       # tags: [team-doc]
```

Currently `note.md`'s effective tags = `[internal-note]` (length-1
arrays merge by index → cascade element at index 0 is overwritten by
`internal-note`). Expected: `[team-doc, internal-note]`.

## Open questions

1. Is replace-semantics intentional for `tags`? Replace makes sense
   for, say, `template.theme` (a scalar wrapped in YAML); it
   surprises for `tags`.
2. If unioning is the fix, is the union scoped to the `tags` key, or
   applied to all keyed-array fields under `frontmatter`/cascade? The
   latter changes the meta-merge contract for fields beyond tags
   (e.g. `pandoc.filters`, `page.headHtml`) and needs its own test
   coverage.
3. Should the fix migrate `mergeAesons` to a tag-aware merger, or
   special-case `tags` at the call site that builds the effective
   metadata?

Found while triaging the review for #352. The #352 fix surfaces this
boundary on the index path; the render path has had it forever.

srid:
**`Emanote.Model.SData.mergeAeson` no longer drops cascade-declared tags when a child note declares its own.** The metadata-cascade merger inherited `aeson-extra`'s `lodashMerge`, which aligns arrays element-by-element — `tags: [team-doc]` in `folder.yaml` plus `tags: [internal-note]` on the leaf produced `[internal-note]`, dropping the cascaded entry both from the per-page chip strip and from the post-#352 global tag index. Closes #697.

### What changed

The merger is rewritten in five lines and now applies one universal contract:

| Value type | Behavior |
|------------|----------|
| Object | Deep-merge by key, recursively |
| Array | Concatenate, then deduplicate (left order preserved) |
| Scalar (or type mismatch) | Right wins |

The array clause is the load-bearing one. _Everything else matches lodash, so the four non-cascade callers (`Note.withAesonDefault`, `Note.overrideAesonText`, `View.Template.setErrorPageMeta`, and `parseSDataCascading` itself) inherit the new behaviour without touching their callsites._

`aeson-extra` was the merger's only user — now removed from `emanote.cabal`.

### Why concatenate-with-dedup is the right cascade semantic

The cascade's job is "defaults that descendants extend." Lodash's array-by-index merge is a third behaviour — neither replace-the-list nor extend-the-list — that comes from JS treating arrays as objects keyed by numeric indices. The same fix generalises to `pandoc.filters` and any other list-valued field a future schema adds; per-key strategies aren't needed.

### Coverage

| Layer | Test |
|-------|------|
| Merge contract | 8 hspec cases in `Emanote.Model.SDataSpec` (object deep-merge, array union, dedup, single-side preservation, nested-array via `pandoc.filters`, scalar override across a 3-step fold) |
| Render path | smoke `/issue-697/note.html` — chip strip carries the cascaded tag plus the leaf's frontmatter and inline-body tags simultaneously |
| Index path | smoke `/-/tags/issue-697-cascaded.html` — lists the cascaded note even when it declares its own tags |

The stale "do NOT add inline `#tags` here" comment on the `issue-352` fixture is gone — the constraint it encoded no longer holds.

### Try it locally

```sh
nix run github:srid/emanote/fix/697-cascade-tags-union
```


srid:

| 1 | Hickey | No structural simplicity violations                                  | No-op             |
| 2 | Lowy   | `mergeAeson` shared by cascade and non-cascade callers — make boundary visible | Fixed in this PR (`8ad010ec`, then trimmed `5cb699f0`) |
| 3 | Lowy   | Per-key merge strategies on the roadmap; current interface stable today | Fixed in this PR — Haddock notes the contract is universal; per-key rules belong in a future module, not a flag here (folded into `8ad010ec`, then trimmed `5cb699f0`) |
| 4 | Lowy   | Cascade vs. non-cascade asymmetry (1 cascade caller, 3 non-cascade)  | Fixed in this PR — Haddock makes the shared contract explicit (folded into `8ad010ec`, then trimmed `5cb699f0`) |

### Hickey rationale

> The implementation is clean. The merge logic is consolidated in one function. The cascading fold cleanly delegates to pairwise merge. Three branches (Object, Array, catch-all scalar) cleanly partition the value space — no hidden state, no temporal coupling. Hand-rolled utility replacing `lodashMerge` is justified: the external library solved the wrong problem (index alignment); the three-line custom implementation is domain-specific and deliberately diverges from the library's array semantics.

### Lowy rationale

> `mergeAeson`/`mergeAesons` are used for three semantically distinct purposes: cascade folding (`Meta.getEffectiveRouteMetaWith`), default application (`Note.withAesonDefault`), and tag union (`Note.overrideAesonText`). All three currently share concatenate-with-dedup semantics for arrays, but they are logically different operations. If a future requirement differentiates them ("suppress cascade for field X"), the boundary will need restructuring. The CHANGELOG noting "different rules per key" as future work makes the current single-function interface worth flagging.

### Disposition notes

- The Lowy reviewer recommended **renaming** `mergeAeson` to `mergeCascadeAeson`. I disagreed and applied the structural concern via Haddock instead: 4 of 5 callers are _not_ cascade, so the cascade-specific name would mislead them more than the current generic name does.
- Two follow-up commits (`5cb699f0`, `2e9dc07c`) trimmed the post-Lowy Haddock and CHANGELOG bullet after the `/code-police` elegance pass flagged disproportionate prose. The Contract section + lodash-divergence paragraph carry the boundary information; the call-site enumeration and stability-note prose were cut as folklore that would rot.

srid:
## Evidence

The page chip strip on the `/issue-697/note.html` fixture renders the cascaded tag alongside the note's own and inline tags. The fixture has `issue-697.yaml` declare `tags: [issue-697-cascaded]` and the leaf note declare `tags: [issue-697-own]` plus an inline `#issue-697-inline`. On `master`, lodash's index-aligned merge clobbers the cascaded entry; with this PR it survives.

| Before (`master`) | After (this PR) |
| --- | --- |
| before | after |

srid:

| research | ✓ | 8m 16s | Identified mergeAeson uses lodashMerge (alignWith for arrays); planned the concat-and-dedup fix |
| branch | ✓ | 14s | Created fix/697-cascade-tags-union from origin/master |
| **implement** | ✓ | **8m 28s** | Rewrote mergeAeson with object-key/array-concat-dedup/scalar-right semantics (drops aeson-extra). 8 hspec cases pass; cabal build all green. New issue-697 fixture + e2e scenarios; refreshed stale issue-352 fixture comment |
| check | ✓ | 21s | cabal build all → up to date |
| docs | ✓ | 36s | Added CHANGELOG bullet and 'Cascade merge semantics' section to docs/guide/yaml-config.md |
| fmt | ✓ | 29s | cabal-fmt, fourmolu, hlint, nixpkgs-fmt all passed |
| commit | ✓ | 54s | Primary fix `575ff9ac` pushed to origin |
| hickey+lowy | ✓ | 4m 42s | Hickey: no findings. Lowy: 3 findings folded into Haddock revision `8ad010ec` |
| police | ✓ | 5m 39s | Pass 1 (rules): clean. Pass 2 (fact-check): clean. Pass 3 (elegance): 2 fixes — `5cb699f0`, `2e9dc07c` |
| test | ✓ | 3m 14s | cabal test all (74/74) + e2e-live (37/37) + e2e-static (35/37, 2 skipped) + e2e-morph (37/37) |
| create-pr | ✓ | 1m 10s | Draft PR opened with hickey/lowy analysis comment |
| ci | ✓ | 3m 8s | vira ci pipeline succeeded on HEAD 2e9dc07c |
| evidence | ✓ | 3m 14s | Before/after chip-strip screenshots posted |
| **Total** | | **40m 50s** | |

### Optimization suggestions

- **Research could shave ~3 min.** Cloning `aeson-extra` to confirm `lodashMerge`'s `alignWith` semantics took most of the research budget. If `Emanote.Model.SData.hs` were pre-read before invoking `/do`, the lodash detour wouldn't be needed — the issue body already names the symbol.
- **Implement+research were the two heaviest steps (~16 min combined, 41% of total)** but neither dominates individually (≥30% threshold not hit). The parallelism that hickey+lowy bought (4m 42s for both) and the `BuildOnly` short-circuit on `vira ci` (3m 8s) kept the back half lean.
- **Police's elegance pass partially walked back the post-Lowy Haddock.** The Lowy reviewer asked for a stability note; the elegance pass cut it as speculative. Future runs could resolve this tension earlier — e.g., by treating Lowy's "future-proofing" recommendations as defers-to-issue rather than fold-into-Haddock by default.
- **Two police commits could have been one.** The Haddock and CHANGELOG trims are independent, so the per-finding rule applies — but a reviewer reading the PR will see four police-flavoured commits and might appreciate them being labelled by category (`refactor(lowy)` vs `refactor(police)`) more consistently.

Workflow completed at 2026-05-04 12:23 UTC.

PR #741, opened 2026-05-17
Hot-reload `.emanoteignore` mid-session (closes #739)

### LINKED ISSUE #739: Hot-reload `.emanoteignore` mid-session
Today, `.emanoteignore` is **read once at process start** — edits during
a `just run` / live-serve session don't take effect until the binary
restarts (e.g. ghcid restarts the program when a `.hs` file changes,
which incidentally re-reads the file).

This is flagged in-code at [`emanote/src/Emanote/Source/Dynamic.hs:66-70`](https://github.com/srid/emanote/blob/master/emanote/src/Emanote/Source/Dynamic.hs#L66-L70):

> Per-layer ignore patterns are loaded once at startup. Edits to a
> `.emanoteignore` file during a live-serve session do not take effect
> until restart. Live reload (#228 phase 2) requires threading an IO
> refresh through unionmount's recursive remount path, which is
> currently a static `Map source [FilePattern]`.

### Why it matters

Live editing of `.emanoteignore` is exactly the moment a developer
wants the fastest feedback — _"why is this folder still showing in the
sidebar?"_ Forcing a restart breaks the flow and contradicts the
no-restart promise of the rest of `emanote run`.

### Where the seam is

`unionmount` (the [`srid/unionmount`](https://github.com/srid/unionmount)
dependency) accepts ignore patterns as a static `Map Loc [FilePattern]`
once, at mount time. Hot reload requires:

1. **Upstream**: a way to swap the per-layer pattern set without
   tearing down the mount. Likely an `IORef`/`TVar` parameter, or a
   reconfiguration callback.
2. **Emanote side**: watch each layer's `.emanoteignore` via the
   existing fsnotify pipeline, re-parse on change, and push the new
   set through the new unionmount entry point. Then walk the model
   and evict/re-include files whose ignore status flipped.

### Scope

- Single-file edit: pattern added → matching notes leave the model and
  the sidebar; pattern removed → previously-ignored notes appear.
- Multi-layer: each layer's `.emanoteignore` is independent — only
  that layer's file-set should re-evaluate.
- File rename/delete: handled by the same fsnotify path.

### Out of scope

- Universal patterns in `Emanote.Source.Pattern.ignorePatterns` (those
  are compiled in by design).

### Workaround until then

Touch any `.hs` file so ghcid restarts emanote, or `Ctrl-C` and
re-run `just run`.

Originally surfaced while moving Ralph build-perf reports under
`docs/dev/` in #738.

srid:
**`emanote run` now picks up edits to a layer's `.emanoteignore` without a restart.** Adding a pattern evicts matching notes and static files from the model and the sidebar; removing a pattern brings the previously-hidden ones back. Closes #739.

The seam the issue described — unionmount accepts ignore patterns as a static `Map source [FilePattern]` — is sidestepped by *moving* per-layer filtering out of unionmount entirely. unionmount keeps only the universal patterns (dotfile-dir contents, vim backups, the reserved `-/`); per-layer `.emanoteignore` content lives in a `TVar` that `.emanoteignore` change events update in place. The streaming handler reads that TVar on every event and the dedicated reload path walks `_modelNotes` + `_modelStaticFiles` whenever the pattern set actually moves.

### Flow

```
fsnotify event
      │
      ▼
unionmount Change ──▶ handle (splits IgnoreFile from rest)
                          │
                          ├── R.IgnoreFile  ─▶  handleIgnoreFileChanges
                          │                       ├── reload patterns from disk
                          │                       ├── evictNewlyIgnored  (notes + static files)
                          │                       └── reEmitModified     (ledger walk)
                          │
                          └── everything else ─▶ applyFiltered ─▶ classifyOverlays
                                                                    ├── Kept     → patchModel as-is
                                                                    ├── Partial  → patchModel with survivors + record
                                                                    └── Dropped  → Delete + record
```

`Emanote.Route.Ext` gains a new `IgnoreFile` constructor so unionmount surfaces `.emanoteignore` events through the same tagged pipeline as `.md`/`.lua`/`.yaml`. The `R.IgnoreFile` branch of `patchModel` is a no-op — by the time an event reaches it, `handleIgnoreFileChanges` has already reshaped the model upstream.

### Coverage

Two new `@live @hot-reload` Cucumber scenarios in `tests/features/emanoteignore.feature` cover both directions:

| Scenario | Setup | Action | Expectation |
|---|---|---|---|
| Removing a pattern re-includes the matching note | Baseline `.emanoteignore` excludes `secret-ignored.md` | Rewrite file without that line | URL `/secret-ignored.html` contains marker within 10s |
| Adding a pattern hides the matching note | Empty `.emanoteignore` | Rewrite to add `secret-ignored.md` | URL stops containing marker within 10s |

### Behavior change

> The previous universal `**/.*/**` pattern silently matched root-level standalone dotfiles too (because filepattern's `**` matches the empty path). Tightening it to `**/.*/**/*` is what lets `.emanoteignore` reach the hot-reload pipeline; the side effect is that root-level `.gitignore`, `.envrc`, `.editorconfig`, etc. — which the old pattern was *incidentally* swallowing — are now exposed as static assets unless the user lists them in their own `.emanoteignore`. *Dotfile directories like `.git/` and `.vscode/` are still filtered, just by the tighter pattern.*

### Notable limitations (documented in [`docs/guide/emanoteignore.md`](../blob/hot-reload-emanoteignore/docs/guide/emanoteignore.md))

YAML data files (the metadata cascade) and Heist `.tpl` templates don't carry per-layer source metadata in the model yet, so adding a pattern matching one mid-session leaves the resolved cascade / template loaded until restart. Removing such a pattern is unaffected because those files were never filtered out in the first place.

### Try it locally

```sh
nix run github:srid/emanote/hot-reload-emanoteignore -- -L ./my-notebook run
```


srid:

| # | Lens   | Finding                                                                                  | Disposition       |
|---|--------|------------------------------------------------------------------------------------------|-------------------|
| 1 | Hickey | `**/.*/**` claim — silently matches root `.emanoteignore`                                | No-op (empirically wrong: filepattern test verified `True/True/True` for the pattern matching root + nested cases. The static "the .emanoteignore file itself is not served" scenario on master also depends on this match working at root.) |
| 2 | Hickey | `evictNewlyIgnored` skips static files                                                   | Fixed in this PR  |
| 3 | Hickey | `_isModifiedEvents` parallel mirror with discipline-only invariants                      | Fixed in this PR  |
| 4 | Hickey | `applyFiltered` complects filter + bookkeeping + dispatch; predicate duplicated          | Fixed in this PR  |
| 5 | Hickey | `isLayerRootIgnoreFile` exported but unused                                              | Fixed in this PR  |
| 6 | Hickey | Session-stable params (`layers`, `noteFn`, `storkIndex`, `IgnoreState`) threaded positionally | Fixed in this PR |
| 1 | Lowy   | `evictNewlyIgnored` is note-only — scope asymmetric to `applyFiltered`                   | Fixed in this PR (subsumed by Hickey #2 + docs caveat for YAML/templates) |
| 2 | Lowy   | `isLayerRootIgnoreFile` dead exported interface                                          | Fixed in this PR (same finding as Hickey #5) |
| 3 | Lowy   | `IgnoreState` / `ModifiedEvent` raw accessors without `makeLenses`                       | Fixed in this PR  |
| 4 | Lowy   | Asymmetric `Patch.IgnoreFile` vs `Dynamic.handleIgnoreFileChanges` interception points undocumented | Fixed in this PR |

### Hickey rationale

The diff initially folded *both* filtering decisions and ledger bookkeeping into single match arms of `applyFiltered` — three `Refresh` cases each pairing a pure outcome (kept / partial / dropped) with a side-effecting record-or-forget plus a dispatch. Lifting the pure decision into `Ignore.classifyOverlays` (returning an `OverlayOutcome` ADT) lets `applyFiltered` and `reEmitModified` read as **classify → bookkeep → dispatch** instead of triple-interleaved per-arm code; the same predicate now lives in one place.

The two TVars (`_isPatterns` + `_isModifiedEvents`) were folded into one `IgnoreSlice` behind a single `TVar`. The discipline that *every* `recordModified` should pair with a matching `forgetModified` is enforced by routing every read through `snapshotSlice` and every write through `updateSlice` — both helpers act on the single slice atomically, so a future invariant linking the two fields has one place to live.

`evictNewlyIgnored` originally only walked `_modelNotes`, leaving static files (PDFs, images, source-code embeds) reachable even after a pattern was added that should hide them. Adding `_staticFileSource :: Maybe (Loc, FilePath)` to `StaticFile` and walking `_modelStaticFiles` alongside notes closes the gap. YAML data and Heist `.tpl` templates carry no source metadata yet — that limitation is documented in the user-facing docs page rather than fixed structurally; both volume and scope made the model-layer refactor disproportionate to the PR's mandate.

Session-stable handler parameters (`layers`, `noteFn`, `storkIndex`, `IgnoreState`) were threaded positionally through five functions; bundled into a `HandlerCtx` record so a new helper that needs (say) `storkIndex` alongside `ignoreState` doesn't grow another positional argument.

### Lowy rationale

The volatility being encapsulated is _"what counts as a layer-root configuration file."_ Master left it overloaded with _"what kinds of source files unionmount routes through the typed pipeline"_ — `.emanoteignore` was filtered universally, with no first-class type. Adding `R.IgnoreFile` to `FileType SourceExt` makes that volatility visible: `Pattern.filePatterns` claims the file (with `IgnoreFile` listed before `AnyExt` so it wins tagging), `Patch.indexesAsStaticFile` excludes it from the static-file route, and `Source.Dynamic.handleIgnoreFileChanges` owns the reshape semantics. A future change to "what counts as a layer-root config file" lands in one place.

The `_sliceModified` ledger encapsulates a volatility _orthogonal_ to the model: the model tracks _resolved_ content, the ledger tracks the _filter decisions_ that produced that content. Re-walking the filesystem on every pattern change would couple the orchestration to filesystem I/O — keeping the modified-event mirror is the right call for this volatility split.

`Emanote.Source.Ignore` (file format + pattern matching) and `Emanote.Source.Dynamic` (mid-session orchestration) are the right two modules for the two volatilities. `isLayerRootIgnoreFile` ended up in Ignore as a name predicate (pure semantics) and is consumed by Dynamic for event routing — clean boundary. The asymmetric interception points between `Patch.patchModel` (LuaFilter hot-reload) and `Dynamic.handleIgnoreFileChanges` (.emanoteignore hot-reload) are structurally justified — different volatilities, different lifecycles — but easy to miss without explicit cross-reference notes, which the diff now adds at both ends.

srid:
## Evidence

Three WebSocket-morphed states captured in a single browser tab at `/secret-ignored` — no page reload between states; the sidebar and content update purely from fsnotify → Ema-WS.

**State 1 — Baseline.** `secret-ignored.md` is listed in `.emanoteignore`. The route renders the "Missing link" error page and the note is absent from the sidebar.

Baseline: note excluded, Missing link error

**State 2 — Pattern removed.** The `secret-ignored.md` line is deleted from `.emanoteignore` on disk. The page morphs in place: the note body appears (marker text `EMANOTEIGNORE_REGRESSION_FILE` is visible) and "Should not appear" is listed in the sidebar.

Pattern removed: note body visible in sidebar

**State 3 — Pattern re-added.** The line is written back to `.emanoteignore`. The page morphs back to "Missing link" and the entry disappears from the sidebar — without restarting the server or reloading the tab.

Pattern re-added: note hidden again

srid:

| sync | ✓ | 0s | forge=github, branch=hot-reload-emanoteignore |
| research | ✓ | 12m 10s | design settled: emanote-only TVar+ledger approach |
| branch | ✓ | 5s | feature branch hot-reload-emanoteignore |
| implement | ✓ | 15m 6s | IgnoreFile FileType; classifyOverlays; TVar-driven hot-reload; 2 cucumber scenarios |
| check | ✓ | 14s | `cabal build all` clean |
| docs | ✓ | 1m 17s | CHANGELOG bullet + emanoteignore.md Hot reload section |
| fmt | ✓ | 49s | hlint + fourmolu + cabal-fmt + nixpkgs-fmt all green |
| commit | ✓ | 24s | primary feature commit pushed |
| **hickey+lowy** | ✓ | **34m 50s** | 10 findings: 1 No-op (empirically incorrect), 7 fix commits, 2 cross-lens dups |
| **police** | ✓ | **28m 9s** | 1 rule fix + 1 fact-check + 7 elegance fixes (9 commits) |
| **test** | ✓ | **39m 56s** | e2e-live 81/81 incl. 2 new scenarios; static 60+21 skipped; unit 125/125 |
| create-pr | ✓ | 1m 32s | draft PR #741 + hickey+lowy analysis comment |
| ci | ✓ | 1m 54s | `vira ci` both signoffs; e2e-morph 81/81 |
| evidence | ✓ | 2m 22s | three before/middle/after screenshots |
| **Total** | | **2h 19m 22s** | |

**Slowest step**: `test` (39m 56s) — dominated by an e2e-live run on a binary built before the fsnotify root-cause was found.

### Optimization suggestions

- **Verify the live happy path manually before running the full e2e suite.** `test` burned ~25 minutes on a green-looking build that the new `@hot-reload` scenarios were silently failing against. A 60-second `cp fixture → emanote run → edit .emanoteignore → curl` would have surfaced the `**/.*/**` filter-pattern bug before the slow Cucumber + Playwright pass.
- **Pre-test filepattern semantics for any pattern you change.** The whole second hour of this run was triggered by `**/.*/**` accidentally matching root-level dotfiles (filepattern's `**` matches the empty path). A 3-line `?==` smoke test would have caught this at design time, in the `research` step.
- **Combine related police fixes that touch the same file.** The seven elegance commits each round-tripped through `nix develop -c cabal build emanote` + `just fmt` (~30s each). Batching the trivial in-file ones (e.g. `lmlRouteFileType` + trim-Haddock + collapse-TVars all touched `Source/Dynamic.hs`) would shave a few minutes off without losing the per-finding commit message clarity.
- **Re-use `--from polish` or `--from ci-only` on retries.** The pattern fix needed only a rebuild + e2e replay, but it walked the full workflow from check onwards. The CLI already supports skipping to the relevant entry point.

Workflow completed at 2026-05-17.

PR #750, opened 2026-06-17
Add support for custom SyntaxMap

TristanCacqueray:
This change integrates https://github.com/srid/heist-extra/pull/19


TristanCacqueray:
Hello @srid , I'm finally catching up with latest emanote. The new syntax highlight is great! Here is a PR to enable adding custom syntax. Not sure if and how to do to it from the config, this presently only works when using emanote as a library. What do you think?

srid:
> Not sure if and how to do to it from the config, this presently only works when using emanote as a library. What do you think?

How are you using this right now?

TristanCacqueray:
Like so: https://codeberg.org/TristanCacqueray/midirus.com/src/branch/main/src/Mitadi.hs#L36

